PR #4, opened 2018-06-17
massive upgrades to everything, weeeee

toddmedema:


toddmedema:
As discussed, merging this in since the main migration craziness is over, we can now switch back to a normal PR flow

PR #8, opened 2018-06-25
update tslint to use recommended & project, --fix

toddmedema:
Points of interest: changes to tslint.json, the general style changes to code.

Also, `lint --fix` is amazing. We can literally have it run pre-commit so that you'll never even get complaints about silly things like import ordering, it's just fixed magically :)


toddmedema:
Builds and runs!

For the sake of upcoming PRs not being a nightmare to merge, going to merge this in and we can continue to refine the style rules (ended up disabling many of the recommended rules for the sake of initial diff sanity)

PR #343, opened 2018-07-03
Quest Creator: GOTO tags linkable to card location

### LINKED ISSUE #334: Quest Creator: GOTO tags linkable to card location
<a href="https://github.com/danielwiktorczyk"></img></a> **Issue by [danielwiktorczyk](https://github.com/danielwiktorczyk)**
_Monday Apr 09, 2018 at 03:43 GMT_
_Originally opened as https://github.com/ExpeditionRPG/expedition-quest-creator/issues/476 (2 comment(s))_

----

Although you could use CTRL+F to find where you stamped the card GOTO ID #townSquare, I was thinking that a clickable link to that card within the quest creator could be a slick and quicker way to travel within the document. 

It would look like this:

```
_Town Square_ (#townSquare)

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut

* Route A 

//------------------------large chunk of code---------------------------

// somewhere along the way:

                            **goto townSquare**

* Route B

//-------------------another large chunk of code-------------------------

// somewhere along this other way, might be 200 lines down:

                            **goto townSquare**
```

In this example, any "goto townSquare" would, upon being hovered over by a mouse, display a clickable link within the document to shoot the user up to the location of the #townSquare.

JTimNolan:
Add functionality for #334, jump to the card ID from a goto command. Hotkey is ctrl/cmd-i with the cursor in the goto token, or alt-click on the goto token. I would have preferred to use ctrl-click, as that is a jump to definition hotkey in other IDEs, but in Ace it seems it's already bound to multi-cursor selection. If we don't need/want that multi-cursor feature (seems like it might not play well with the editor preview?) I'd recommend we switch to using ctrl-click for this.

This is my first time using TypeScript, first time with Ace, first time contributing to this repo; lots of first times so please let me know if some revisions are needed. On that note, I'm not really sure how to write a test for this if one is required. TextView.test.tsx is empty to I couldn't find a good example. If you can point me to one I'm happy to oblige.

Closes #334


services/quests/src/components/base/TextView.tsx
```diff
@@ -135,6 +136,41 @@ export default class TextView extends React.Component<TextViewProps, {}> {
...
+        if (!e.domEvent.altKey) {
+          return;
+        }
+        ref.editor.commands.commands.jumpToDef.exec();
+      });
+
+      ref.editor.commands.addCommand({
+        name: 'jumpToDef',
+        hint: 'Jump to defintion',
+        bindKey: {win: 'Ctrl-I',  mac: 'Command-I'},
```
toddmedema: Overall this looks great! Only thing I can think of would be to change this to ctrl/cmd - J for "jump to"

JTimNolan: I thought the same thing, but in Chrome ctrl-j is the shortcut to open the downloads page. I think Ace is supposed to stop those events from bubbling, but while I was testing it didn't seem to do so consistently. Seemed better to me to just go with something that isn't a conflict.

toddmedema: Makes sense to me, thank you for clarifying!

toddmedema:
Do the tests pass on your local environment?

There's an unfortunate known issue with Travis (our CI tool) where they don't allow PRs from outsiders to access the environment variables for security reasons (which is reasonable). That being said, our tests shouldn't use / depend on private variables / API keys, so I've opened https://github.com/ExpeditionRPG/expedition/issues/348 if you're interested in taking a crack at it!

JTimNolan:
No, as I understand it I need to set up a GCP project and mimic your postgres environment in order for them to work. I haven't set that up yet, so I get the same errors on my side. But I was hoping that Travis would be configured to have those tests work.

toddmedema:
Fair :) Just pulled this down and confirmed that the tests work. Only one change - can you update it to jump to the line, instead of the line + 1? It'll give users more context to see the card title and ID that they just jumped to :)

In the mean time, I'll work on getting CI tests up and running as expected.

Thank you, and congratulations on your first contribution! Do you have another issue picked out that you'd like to work on?

(also, cool github trick - if you say "Closes #334" on a line in your PR description, when the PR is merged in it'll auto-close the issue!)

JTimNolan:
Oh nice, I'll try that out as well!

JTimNolan:
Just tried it locally, seems to bypass the tests it was failing before, but now it fails many more as a result. I'll push to this branch so you can get the log

JTimNolan:
Ok... nevermind :) I'll see if I can sort out the issue on my end!

toddmedema:
Awesome! Since you just went through that installation and the steps are fresh in your head, can you create a PR to update the README to help new folks?

toddmedema:
(also, if you can email me at Todd@Fabricate.io with the email address your Expedition account is under, I'll make sure to get your account credited with those loot points! :)

toddmedema:
Hey @JTimNolan - Hadn't heard back from you regarding your email so that we can get you those loot points :) There's also several other open issues that I think you'd enjoy!

Issue #258, opened 2018-07-01
Use standardized QuestListElement in saved quests + featured quests


toddmedema:
<a href="https://github.com/smartin015"><img src="https://avatars2.githubusercontent.com/u/607666?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [smartin015](https://github.com/smartin015)**
_Friday Jun 01, 2018 at 00:19 GMT_
_Originally opened as https://github.com/ExpeditionRPG/expedition-app/issues/719 (1 comment(s))_

----

Aka we'll want to split this (and its tests) into its own components so that we can use it in multiple places - https://github.com/ExpeditionRPG/expedition-app/blob/master/app/components/views/Search.tsx#L305

toddmedema:
<a href="https://github.com/toddmedema"><img src="https://avatars2.githubusercontent.com/u/775657?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [toddmedema](https://github.com/toddmedema)**
_Monday Jun 04, 2018 at 23:40 GMT_

----

@ezittel this would be another good practice PR after the paper and pen update :)

Issue #242, opened 2018-07-01
Favoriting


toddmedema:
<a href="https://github.com/toddmedema"><img src="https://avatars2.githubusercontent.com/u/775657?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [toddmedema](https://github.com/toddmedema)**
_Sunday Apr 29, 2018 at 04:43 GMT_
_Originally opened as https://github.com/ExpeditionRPG/expedition-app/issues/668 (0 comment(s))_

----

So, we want to show users two new cards / way of going through quests: their quest history and favorites.

API: enable the quest search endpoint to support being passed an array of IDs; the app then passes in the IDs for all of the user's history or favorites quests, and we render those cards as effectively another search page.

To support favoriting, we'll also need /favorite and /unfavorite endpoints, and we'll want to show the favorite icon in results cards + quest details.

toddmedema:
Now that we have quest history, this is less valuable

PR #347, opened 2018-07-05
Quest history page & additional fixes

smartin015:
https://github.com/ExpeditionRPG/expedition/issues/242

Adds a "Quest history" beta feature - the API server now queries for quest details using the analytics events for quest completion, returning it to the app which then has a flow similar to SavedQuests for interaction.

Additional fixes:
- Add `yarn app-watch-test` and `yarn api-watch-test` for continuous testing
- Analytics events weren't being saved on quest begin and end (it wasn't dispatching the thunk). Fixed this and added tests
- Moved the quest history out of the `user` reducer, since it's not really a property of the user and we may extend it later on.

### DIFF
--- a/services/api/src/models/Users.test.ts
+++ b/services/api/src/models/Users.test.ts
@@ -1,9 +1,9 @@
 import {AnalyticsEvent} from 'shared/schema/AnalyticsEvents';
 import {Database} from './Database';
 import {analyticsEvents as ae,
+  quests as q,
   testingDBWithState,
-  users as u,
-  quests as q
+  users as u
 } from './TestData';
 import {
   getUser,
--- a/services/api/src/models/Users.test.ts
+++ b/services/api/src/models/Users.test.ts
@@ -2,7 +2,8 @@ import {AnalyticsEvent} from 'shared/schema/AnalyticsEvents';
 import {Database} from './Database';
 import {analyticsEvents as ae,
   testingDBWithState,
-  users as u
+  users as u,
+  quests as q
 } from './TestData';
 import {
   getUser,
@@ -73,9 +74,10 @@ describe('users', () => {
       .catch(done.fail);
     });
 
-    it('returns valid results for players with quest history', (done: DoneFn) => {
+    fit('returns valid results for players with quest history', (done: DoneFn) => {
       testingDBWithState([
         u.basic,
+        q.basic,
         new AnalyticsEvent({...ae.questEnd, userID: u.basic.id}),
       ])
       .then((tdb) => {
@@ -85,6 +87,7 @@ describe('users', () => {
         expect(Object.keys(quests).length).toEqual(1);
         const result = quests[ae.questEnd.questID];
         expect(result.lastPlayed).toEqual(ae.questEnd.created);
+        expect(result.details.title).toEqual(q.basic.title);
         done();
       }).catch(done.fail);
     });
--- a/services/api/src/models/Users.test.ts
+++ b/services/api/src/models/Users.test.ts
@@ -74,7 +74,7 @@ describe('users', () => {
       .catch(done.fail);
     });
 
-    fit('returns valid results for players with quest history', (done: DoneFn) => {
+    it('returns valid results for players with quest history', (done: DoneFn) => {
       testingDBWithState([
         u.basic,
         q.basic,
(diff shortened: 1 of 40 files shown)


services/api/src/models/Users.test.ts
```diff
@@ -73,9 +74,10 @@ describe('users', () => {
       .catch(done.fail);
     });
 
-    it('returns valid results for players with quest history', (done: DoneFn) => {
+    fit('returns valid results for players with quest history', (done: DoneFn) => {
```
toddmedema: typo with the `f`?

services/api/src/models/Users.ts
```diff
@@ -57,6 +60,27 @@ export function getUserQuests(db: Database, id: string): Bluebird<UserQuestsType
         lastPlayed: new Date(result.get('lastPlayed')),
       };
     });
-    return userQuests;
+
+    const metas: Bluebird<void>[] = [];
+    for (const k of Object.keys(userQuests)) {
+      metas.push(db.quests.findOne({
```
toddmedema: This does a separate DB lookup for each quest the user has played? Can we do a join here so that this isn't an n^2 lookup?

services/app/src/components/views/QuestHistory.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+export interface QuestHistoryDispatchProps {
+  onSelect: (selected: UserQuestInstance) => void;
+  onPlay: (details: QuestDetails) => void;
+  onReturn: () => any;
+}
+
+export interface QuestHistoryProps extends QuestHistoryStateProps, QuestHistoryDispatchProps {}
+
+function renderDetails(props: QuestHistoryProps): JSX.Element {
```
toddmedema: Is this identical to the search renderDetails function? If we're making another duplicate, can we instead de-duplicate it and close https://github.com/ExpeditionRPG/expedition/issues/258 ?

smartin015: Nope, there's several areas that are slightly different. Might make sense for this to become a proper react component eventually, though.

services/api/src/models/Users.test.ts
smartin015: Good catch - `fit` actually restricts the test to only the unit tests with the `f` prefix... was using it to debug that particular line. Removed.

smartin015:
Users aren't ephemeral - in fact, there's no affordance at all for a user to be able to log out of the app currently.

I always thought of the `user` reducer as more of controlling auth state ("I am this user"), not about information related to a user ("The user has this stuff").

toddmedema:
Hmm. So, there's a fair amount of info "about" the user and that amount is going to continue to grow - their loot points, name, questing history, (and eventually favorites, perhaps credit card information, recommendations, friends, etc). The question in my mind, then, is how we'd want to architect it going forward, either having all of those pieces of information in separate stores or under user.

services/app/src/actions/Card.tsx
```diff
@@ -40,6 +41,12 @@ export const toCard = remoteify(function toCard(a: ToCardArgs, dispatch: Redux.D
     keylist.push(a.keySuffix);
   }
 
+  // Log the end of the quest if we reach it.
+  if (a.name === 'QUEST_END') {
+    console.log('Dispatching logQuestPlay');
+    dispatch(logQuestPlay({phase: 'end'}));
```
toddmedema: Out of curiosity, why did this get moved from the Quest End card to here? And, if so, would it make sense to also move the log start of play dispatch to here as well, to keep all of our card / navigation-based analytics in one spot?

services/app/src/actions/Quest.tsx
```diff
@@ -75,3 +77,15 @@ export function loadNode(node: ParserNode, details?: QuestDetails) {
     }
   };
 }
+
+export function userQuestsDelta(delta: Partial<UserQuestsType>) {
```
toddmedema: Shouldn't these be in `/actions/QuestHistory`?

smartin015: True that. Fix'd.

services/app/src/actions/Card.tsx
smartin015: It was moved to here because logQuestPlay is a thunk action, requiring a call to `dispatch`... having it be part of the component constructor is really weird, since that should just be a view and not dispatch actions directly.

But it's a bit sloppy to have it here in `toCard` as well... I moved this to a separate `endQuest` action which seems a bit nicer.

PR #352, opened 2018-07-06
Consolidated quest preview code, plus small fixes

smartin015:
- There's now a "QuestPreview" component and container that shows quest metadata. This dedupes similar logic in saved quests, quest history, featured quests, and search.
- Quest history didn't persisting when a user selected "Home" because it was counted as back-button-able... fixed this.
- FeaturedQuests interface had several dispatchers that called toCard under the hood; exposed a toCard dispatch function to reduce boilerplate.
- Redux complained about buttons within buttons when we used the StarRating component inside of a clickable quest list item; made the buttons render in a div when they're not interactive to fix this.
- Redux complained about list items when showing requirements for quests (e.g. pen & paper); I todo'd the todo comment on making this logic better.

### DIFF
--- a/services/app/src/components/base/StarRating.tsx
+++ b/services/app/src/components/base/StarRating.tsx
@@ -37,17 +37,28 @@ export default class StarRating extends React.Component<StarRatingProps, {}> {
       }
 
       // TODO: Flatten this structure so ripples are circular
-      return <div key={i} className={classes.join(' ')}>
-        <IconButton
-          disabled={this.props.readOnly}
-          onClick={() => {
-            if (!this.props.readOnly && this.props.onChange) {
-              this.props.onChange(i);
-            }
-          }}>
-          {star}
-        </IconButton>
-      </div>;
+      // We use a non-button element here due to redux errors on nesting buttons
+      // when the stars are just used in a display-only fashion.
+      if (this.props.readOnly) {
+        return (
+          <div key={i} className={classes.join(' ')}>
+            {star}
+          </div>
+        );
+      } else {
+        return (
+          <div key={i} className={classes.join(' ')}>
+            <IconButton
+              onClick={() => {
+                if (this.props.onChange) {
+                  this.props.onChange(i);
+                }
+              }}>
+              {star}
+            </IconButton>
+          </div>
+        );
+      }
     });
     return <span className="starContainer">
       <div className="stars" style={this.props.style}>
--- a/services/app/src/components/views/FeaturedQuestsContainer.tsx
+++ b/services/app/src/components/views/FeaturedQuestsContainer.tsx
@@ -2,11 +2,12 @@ import {connect} from 'react-redux';
 import Redux from 'redux';
 
 import {toCard} from '../../actions/Card';
-import {search, viewQuest} from '../../actions/Search';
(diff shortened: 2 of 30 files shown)


toddmedema:
Does this close https://github.com/ExpeditionRPG/expedition/issues/258 ?

services/app/src/components/base/StarRating.tsx
```diff
@@ -37,17 +37,28 @@ export default class StarRating extends React.Component<StarRatingProps, {}> {
...
-            if (!this.props.readOnly && this.props.onChange) {
-              this.props.onChange(i);
-            }
-          }}>
-          {star}
-        </IconButton>
-      </div>;
+      // We use a non-button element here due to redux errors on nesting buttons
+      // when the stars are just used in a display-only fashion.
+      if (this.props.readOnly) {
```
toddmedema: Out of curiosity, does this fix the alignment bug with the stars vs review count in search result list items?

services/app/src/components/views/FeaturedQuestsContainer.tsx
```diff
@@ -19,14 +20,8 @@ const mapStateToProps = (state: AppState, ownProps: FeaturedQuestsStateProps): F
...
   return {
-    onTools(): void {
-      dispatch(toCard({name: 'ADVANCED'}));
-    },
-    onSavedQuests(): void {
-      dispatch(toCard({name: 'SAVED_QUESTS', phase: 'LIST'}));
-    },
-    onQuestHistory(): void {
-      dispatch(toCard({name: 'QUEST_HISTORY', phase: 'LIST'}));
+    toCard(name: CardName): void {
```
toddmedema: Beautiful!

services/app/src/components/views/QuestHistory.tsx
```diff
@@ -1,55 +1,22 @@
...
-      </div>
-    </Card>
-  );
-}
-
-function renderList(props: QuestHistoryProps): JSX.Element {
+const QuestHistory = (props: QuestHistoryProps): JSX.Element => {
   if (Object.keys(props.played).length === 0) {
     return (
       <Card title="Quest History">
```
toddmedema: Question re: Quest History: What happens if you have multiple saves of the same quest ID? Want to make sure that flow works :) (that you're able to play both saves, that both saves correctly show their saved time, etc)

services/app/src/components/base/StarRating.tsx
smartin015: It does indeed :)

services/app/src/components/views/QuestHistory.tsx
smartin015: It correctly propagates the save timestamp through to the quest preview and plays the quest save specified by the user.

FYI I'm working on a derivative PR now that actually turns the "Saved Quests" list into a "Saved & Offline Quests" list, where each item is a distinct quest that shows the number of saves (and if it's playable from the beginning while offline). Clicking through will go to the consolidated preview page and will list the saved instances of that quest as part of the preview... which means we'll also get this for free on e.g. previews from the search or featured quests pages.

smartin015:
Re: #258 - it doesn't fix it... but we should do something about that eventually ;)

PR #353, opened 2018-07-06
Offline Quests

smartin015:
- Adds an icon to search results indicating quests that have been saved offline
- Adds an icon to the preview page indicating the quest is available offline
- Add options for users to store quests offline/remove the offline copy from the preview page
- Adds a "continue from latest save" button to the preview page
- Adds a time-ordered list of saves for the given quest on the preview page
- Renames "Saved Quests" card to "Saved & Offline" and now also shows offline quests
- Clusters saves in the "Saved & Offline" list by quest, showing the most recent save time and the number of saves
- Move standard formatting helper methods to `Format.tsx` so they can be shared with minimal dependencies
- Add a `pluralize` formatting method that naively gloms on an 's' to nouns that should be made plural
- This extends the localstorage schema with a `pathLen` attribute indicating how far into the quest a user went before they saved. This info is listed on the preview page next to each save, and is used to detect whether a given instance is a "user save during a quest" or "saving a quest for offline play". This also allows for `pathLen=undefined` so as not to break users that enabled experimental features.

offline_quests_demo

### DIFF
--- a/services/app/src/Format.test.tsx
+++ b/services/app/src/Format.test.tsx
@@ -1,4 +1,4 @@
-import {formatPlayPeriod} from './Format';
+import {formatPlayPeriod, smartTruncateSummary, pluralize} from './Format';
 
 describe('formatPlayPeriod', () => {
   it('formats time ranges to minutes and hours', () => {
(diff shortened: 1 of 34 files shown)


services/app/src/Format.test.tsx
```diff
@@ -8,3 +8,32 @@ describe('formatPlayPeriod', () => {
...
+    expect(smartTruncateSummary(SHARDS_OF_TIME_SUMMARY)).toEqual(SHARDS_OF_TIME_EXPECTED);
+  });
+
+  it('leaves excessively-long sentences alone', () => {
+    const MUNROE_SUMMARY = 'This kid-friendly, spooky Halloween adventure takes you into Mr Monroe’s haunted mansion where unexplainable things are happening…';
+    expect(smartTruncateSummary(MUNROE_SUMMARY)).toEqual(MUNROE_SUMMARY);
+  });
+});
+
+describe('pluralize', () => {
```
toddmedema: Can we switch to using https://github.com/blakeembrey/pluralize, which also supports irregular pluralizations and singularizations, and has no extra dependencies / doesn't bloat the package :)

smartin015: Done.

PR #355, opened 2018-07-07
consolidate app deploy scripts

toddmedema:
In preparation for having a root-level deploy command that lets you select your service -> its deploy script.

Can you pull this down on your computer to confirm that the bash script runs as expected on Linux? (note that since the Webpack PR was merged to master, you'll also need to update your global webpack npm module to get it to work)


services/app/deploy.sh
```diff
@@ -0,0 +1,93 @@
...
+      ./zipalign -v 4 platforms/android/app/build/outputs/apk/release/app-release-unsigned.apk platforms/android/app/build/outputs/apk/release/expedition.apk
+
+      # iOS
+      cordova build ios
+
+      # Deploy web app to prod with 1 day cache for most files, 6 month cache for art assets
+      export AWS_DEFAULT_REGION='us-east-2'
+      aws s3 cp www s3://app.expeditiongame.com --recursive --exclude '*.mp3' --exclude '*.jpg' --exclude '*.png' --cache-control max-age=86400 --cache-control public
+      aws s3 cp www s3://app.expeditiongame.com --recursive --exclude '*' --include '*.mp3' --include '*.jpg' --include '*.png' --cache-control max-age=15552000 --cache-control public
+      aws s3 cp s3://app.expeditiongame.com/expedition.apk s3://app.expeditiongame.com/apk-archive/expedition-$version.apk --cache-control public
```
smartin015: AFAICT this copies the previous version of the app and labels it as the current version...

services/app/deploy.sh
```diff
@@ -0,0 +1,93 @@
...
+
+function prebuild() {
+  # clear out old build files to prevent conflicts
+  rm -rf www
+  rm platforms/android/app/build/outputs/apk/debug/app-debug.apk
+  rm platforms/android/app/build/outputs/apk/release/expedition.apk
+}
+
+init
+echo "Where would you like to deploy the app? Current version: ${version}"
```
smartin015: This just echoes and doesn't seek user input... won't it just try to deploy to beta, then to prod?

services/app/deploy.sh
```diff
@@ -0,0 +1,93 @@
...
+OPTIONS="Beta Prod"
+select opt in $OPTIONS; do
+  if [ "$opt" = "Beta" ]; then
```
smartin015: Suggest making these separate functions for readability.

toddmedema: Good idea, fixed

PR #356, opened 2018-07-07
Simplify render pathway for QDL templates 

smartin015:
This removes a lot of references to "combat" and "roleplay" in the QDL render modules, in preparation for adding a "decision" template for time-based skill checks within a quest.

This comes with two small behavior tweaks, the first of which might throw errors on older quests with poor QDL style:

1. Some whitespace is now mandatory between the enemy list and the first event bullet in combat nodes.
1. Roleplay cards titled `roleplay` will have no title.


smartin015:
Added an explicit error about whitespace issues.

I'm less convinced it's worth the effort to have an error on title=roleplay, since it's basically the behavior of `combat` (and soon, `decision`) to replace the title with the type of template being evaluated. If someone names something exactly `_roleplay_` and complains about the lack of a title, you can have an I-told-you-so :)

smartin015:
Done

PR #362, opened 2018-07-08
Add decision QDL template

smartin015:
- Upgrades QC error testing to make one test case per success/failure instance
- Move template-handling code into `schema/` and out of QDL. Base QDL rendering/parsing code is now template-free!
- Adds the new "decision" template type.

Logic for decision validation:
- When a user makes a `_decision_`, they must configure at least 3 skill checks, with viable success and failure conditions for each check.
- Skill checks are either a skill (knowledge/charisma/athletics) OR a skill & persona type ("light knowledge", "dark charisma", etc).
- When users specify events in a decision node, they can specify any of persona+skill+outcome (fully specified), persona+skill (outcome is assumed 'success'), skill (outcome assumed success, applies to persona+skill if exists or otherwise becomes its own check), skill+outcome (applies to persona+skill if exists or otherwise becomes its own check) or outcome (which applies to all checks).
- Implementation is expected whereby the most specific event is chosen matching the in-game outcome. If there are two of equal specificity, one may be chosen at random.


shared/schema/templates/Decision.tsx
```diff
@@ -0,0 +1,145 @@
...
+export enum Skill {
+  ATHLETICS = 'athletics',
+  KNOWLEDGE = 'knowledge',
+  CHARISMA = 'charisma',
+}
+export enum Outcome {
+  SUCCESS = 'success',
+  FAILURE = 'failure',
+  RETRY = 'retry',
+  INTERRUPTED = 'interrupted',
```
toddmedema: Interrupted feels like there could be a better / shorter word for it if writers will be typing it a lot - miss? Guess it's weird since failure can mean either "failure with penalties" like crit fail, or "failure to succeed" like miss

shared/schema/templates/Decision.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+    it('allows for generic checks', () => {
+      expect(getPossibleChecks([{persona: undefined, skill: 'knowledge'}])).toEqual(['knowledge']);
+    });
+    it ('uses only specific checks where possible', () => {
+      expect(getPossibleChecks([
+        {persona: undefined, skill: 'knowledge'},
+        {persona: 'dark', skill: 'knowledge'},
+      ])).toEqual(['dark knowledge']);
+    });
+    it ('uses only specific checks where possible', () => {
```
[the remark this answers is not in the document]
smartin015: Removed

shared/schema/templates/Decision.tsx
smartin015: Let's think about this and change it later.

smartin015:
Re: conditional checks - we get it for free with generic QDL templates and node evaluation... worth ensuring this works when we implement app-side logic though.

PR #376, opened 2018-07-13
Update package commands with start-dev script for file watching

JTimNolan:
Just a set of scripts with the --hot flag to hotreload the page when the file is saved. I usually prefer to develop with that setup so I figured I'd share :) . In the future if we have other improvements for the development environment, we can add them to the -dev scripts as well.


JTimNolan:
Two reasons:
1. Not sure what your production set up is. If you use `yarn run start` in prod, you don't want to have file watchers active for multiple reasons.
2. Didn't want to be presumptuous, maybe some people don't to use hotloading?

If these aren't issues then I agree about consolidating them

toddmedema:
Good calls on both!

1. Nope, everything is built & deployed via `yarn run build`. That being said, the way API is built onto the server is a bit weird, and API development currently doesn't do hot reloading, so we should skip this for API specifically.
2. Appreciate it :)

toddmedema:
What're you thinking of tackling next?

PR #412, opened 2018-08-03
Check if MathJS block should be conditional

### LINKED ISSUE #63: Bug with multiple lines of ops getting evaluated as a conditional paragraph
_Wednesday Jan 31, 2018 at 01:17 GMT_
_Originally opened as https://github.com/ExpeditionRPG/expedition-qdl/issues/59_

----

Ran into this edge case yet again:
```
{{ horrorExpansion = false }}
{{ demandGold = false }}
{{ hagClue = 0 }}
{{ townAnxiety = 3 }}
{{ dayHours = 0 }}
{{ gold = 0 }}
```

JTimNolan:
Closes #63 . Fix is based on the method we discussed @toddmedema , using MathJS's AST API to parse each block ahead of time and determine if it should be conditional. Surprisingly that entire check seemed to fall under simply checking if the outermost node was an OperatorNode - this almost seems too good to be true to be honest, but as far as I can tell it does the trick.

I added some tests for 'normal' cases, idk how far we want to go on testing this, but I can add more if needed.

### DIFF
--- a/shared/render/render/XMLRenderer.test.tsx
+++ b/shared/render/render/XMLRenderer.test.tsx
@@ -35,6 +35,25 @@ describe('XMLRenderer', () => {
     });
 
     it('renders decision'); // TODO
+
+    it('renders conditional ops as ifs', () => {
+      expect(XMLRenderer.toTemplate('roleplay', {}, ['{{ gold == 0 }} Test'], 0).toString())
+        .toEqual('<roleplay data-line="0"><p if=" gold == 0 "> Test</p></roleplay>');
+    });
+
+    it('renders nonconditional ops as output', () => {  
+      expect(XMLRenderer.toTemplate('roleplay', {}, ['{{ gold }}'], 0).toString())
+      .toEqual('<roleplay data-line="0"><p>{{ gold }}</p></roleplay>');
+
+      expect(XMLRenderer.toTemplate('roleplay', {}, ['{{ gold }} Test'], 0).toString())
+        .toEqual('<roleplay data-line="0"><p>{{ gold }} Test</p></roleplay>');
+        
+      expect(XMLRenderer.toTemplate('roleplay', {}, ['{{ gold }}', 'Test on a new line'], 0).toString())
+      .toEqual('<roleplay data-line="0"><p>{{ gold }} Test on a new line</p></roleplay>');
+      
+      expect(XMLRenderer.toTemplate('roleplay', {}, ['{{ test = false }}', '{{ gold = 10 }}'], 0).toString())
+        .toEqual('<roleplay data-line="0"><p>{{ test = false }} {{ gold = 10 }}</p></roleplay>');
+    });
   });
 
   describe('toTrigger', () => {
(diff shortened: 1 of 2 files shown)


shared/render/render/XMLRenderer.tsx
```diff
@@ -49,8 +50,11 @@ export const XMLRenderer: Renderer = {
         const visible = (text.match(INITIAL_OP_WITH_PARAGRAPH) || [])[1];
         let paragraph = `<p>${sanitizeStyles(text)}</p>`;
         if (visible) {
-          text = text.replace('{{' + visible + '}}', '');
-          paragraph = `<p if="${escapeXml(visible)}">${sanitizeStyles(text)}</p>`;
+          const visibleTree = Math.parse(visible);
+          if (visibleTree.type === 'OperatorNode') {
```
toddmedema: Would be great to add a comment here describing what's going on since it's a bit magical :)

shared/render/render/XMLRenderer.test.tsx
```diff
@@ -35,6 +35,25 @@ describe('XMLRenderer', () => {
     });
 
     it('renders decision'); // TODO
+
+    it('renders conditional ops as ifs', () => {
+      expect(XMLRenderer.toTemplate('roleplay', {}, ['{{ gold == 0 }} Test'], 0).toString())
+        .toEqual('<roleplay data-line="0"><p if=" gold == 0 "> Test</p></roleplay>');
+    });
+
+    it('renders nonconditional ops as output', () => {
```
toddmedema: Could you add a few more test cases, just to make absolutely sure this'll work as expected? Few I can think of:
- ternary operator
- assigning something to a dictionary, ie `a = {b: 'c'}`
- accessing something from an array, ie `a = [1,2,3]` then later on outputting `a[1]`
- conditional involving a dictionary

toddmedema:
@JTimNolan looks like one of the tests is failing :(

JTimNolan:
I have encountered one sort of bad case with this solution. I doubt anybody would run into this and in some ways it's kind of their fault if they do, but here it is:

```
{{ assign = {a: 1, b:2}; a > 1 }} I want to both keep assign and only show this text if a>1
```
The obvious work around for this is just doing
```
{{ assign = {a: 1, b:2} }}
{{ a > 1 }} I want to both keep assign and only show this text if a>1
```
but I figured I'd mention that this update won't "fix" that case (because if it's an IF, the assignment is lost)

JTimNolan:
@toddmedema Yep I'm fixing up the tests now!

toddmedema:
Agreed - in fact, it would be nice if we could detect and throw an error in the quest creator if someone tries to do that - https://github.com/ExpeditionRPG/expedition/tree/master/services/quests/errors/definitions - (aka that only one operation per op node is allowed)

JTimNolan:
@toddmedema I was thinking that too - can you point me to an example of how to throw an error in the quest editor?

toddmedema:
Something like this (in the relevant spot in the code) to log the error - https://github.com/ExpeditionRPG/expedition/blob/master/shared/render/render/BlockRenderer.tsx#L37

And then a matching new error code file. I'd recommend setting up the error code first and running tests to make sure the new error definition file is getting picked up.

JTimNolan:
Thanks, I'll check it out

JTimNolan:
Sounds good. Thanks for the feedback!

Issue #399, opened 2018-07-28
Error: Could not get next node for event win


smartin015:
Multiple users are reporting feedback on this error. example comment: "Just finished a battle, moving past the battle result screen"

Happens on these quests:
- The Dread Apprentice #1
- Splice ("End of combat With the skeleton mage the adventure Wont continue")
- The Burnham Terror ("two Nightgaunts, a Nightblade, and an Acolyte, we hit Next and the app crashed")

PR #434, opened 2018-08-31
Improve mid-combat roleplay code.

smartin015:
- Now allows for `**win**` and `**lose**` trigger elements in QDL to trigger the win and lose events. This is a departure from the previous behavior, where `**goto ID**` would show the win or lose screen when the target was within the win or lose event of combat.
- Actually stop combat audio on edge cases (e.g. `**end**` trigger within combat RP)
- Fixes `yarn app-watch-test` so it actually rebuilds and executes the tests on file change
- Add some safety code to stop proliferation of "unknown event win/lose" at the end of combat. If we reach the victory card with a parsernode that's a roleplay node *within* the combat node, it correctly chooses the next node based on the parent combat node.
- resolve phase was improperly setting quest node before taking a history snapshot, which may have contributed to bugs.

#399

### DIFF
--- a/services/app/karma.conf.js
+++ b/services/app/karma.conf.js
@@ -1,31 +1,30 @@
 const webpackConfig = require('./webpack.config');
-const webpack = require('webpack');
 
 webpackConfig.module.rules.unshift({
   test: /isIterable/,
   loader: 'imports?Symbol=>false'
 });
+webpackConfig.entry = undefined;
+
+// Remove hot reload
+webpackConfig.plugins = [webpackConfig.plugins[webpackConfig.plugins.length-1]];
 
 module.exports = function(config) {
   config.set({
     basePath: '',
     frameworks: ['jasmine'],
     files: [
-      { pattern: 'src/**/*.test.tsx' }
+      // Set watched=false here as karma-webpack does the watching under the hood.
+      { pattern: 'src/**/*.test.tsx', watched:false},
     ],
     preprocessors: {
       'src/**/*.test.tsx': ['webpack'],
     },
-    webpack: {
-      module: webpackConfig.module,
-      resolve: webpackConfig.resolve,
-      node: webpackConfig.node,
-      // Pull in module-specific configs (esp. tslint)
-      plugins: [webpackConfig.plugins[webpackConfig.plugins.length-1]],
-      externals: {
-        'react/addons': true,
-        'react/lib/ExecutionEnvironment': true,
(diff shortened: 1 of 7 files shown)


services/app/src/components/views/quest/cardtemplates/combat/Actions.test.tsx
```diff
@@ -172,20 +173,20 @@ describe('Combat actions', () => {
 
     it('randomly assigns damage', () => {
       const {actions} = newStore({});
-      expect(actions[3].node.ctx.templates.combat.mostRecentAttack.damage).toBeDefined();
+      console.log(actions);
```
toddmedema: rm console

services/app/karma.conf.js
```diff
@@ -1,31 +1,30 @@
...
-const webpack = require('webpack');
 
 webpackConfig.module.rules.unshift({
   test: /isIterable/,
   loader: 'imports?Symbol=>false'
 });
+webpackConfig.entry = undefined;
+
+// Remove hot reload
+webpackConfig.plugins = [webpackConfig.plugins[webpackConfig.plugins.length-1]];
```
toddmedema: Would it be possible to be more principled about this, ie to explicitly remove the hot reload plugin rather than assume it's position?

services/app/src/components/views/quest/cardtemplates/combat/Actions.test.tsx
```diff
@@ -239,6 +240,32 @@ describe('Combat actions', () => {
 
       expect(store.getActions()[1].node.ctx.templates.combat.levelUp).toEqual(false);
     });
+
+    it('Fixes when given a mid-combat roleplay node', () => {
```
toddmedema: This test title feels unclear what it's actually testing?

services/app/src/components/views/quest/cardtemplates/combat/Actions.test.tsx
smartin015: Done

services/app/src/components/views/quest/cardtemplates/combat/Actions.test.tsx
smartin015: Done.

services/app/karma.conf.js
smartin015: Made an attempt at this :) hard to do when the copy plugin has poor naming though.

PR #437, opened 2018-09-01
add The Future to API, QC, app

toddmedema:


services/api/src/models/Quests.test.ts
```diff
@@ -10,7 +10,7 @@ import {
 describe('quest', () => {
   describe('searchQuests', () => {
     it('returns an empty array if no results', (done: DoneFn) => {
-      testingDBWithState([q.basic, q.expansion])
+      testingDBWithState([q.basic, q.horror, q.future])
```
smartin015: Maybe this array should be a `const` value in the root of the `searchQuests` section...

services/api/src/models/Quests.ts
```diff
@@ -128,7 +134,8 @@ function mailNewQuestToAdmin(mail: MailService, quest: Quest) {
     over ${quest.mintimeminutes} - ${quest.maxtimeminutes} minutes.
     ${quest.genre}.
     ${quest.requirespenpaper ? 'Requires pen and paper.' : 'No pen or paper required.'}
-    ${quest.expansionhorror ? 'Requires The Horror expansion.' : 'No expansions required.'}`;
+    ${quest.expansionhorror ? 'Requires The Horror expansion.' : 'No expansions required.'}
+    ${quest.expansionfuture ? 'Request The Future expansion.' : ''}`;
```
smartin015: Is this intentionally different?

services/app/src/Constants.tsx
```diff
@@ -26,6 +26,8 @@ export const MULTIPLAYER_SETTINGS = {
 export const FEATURED_QUESTS: QuestDetails[] = [ // Featured quest ids generated from publishing, but don't leave them published!
   {id: '0B7K9abSH1xEOeEZSVVMwNHNqaFE', partition: 'expedition-public', theme: 'base', official: true, title: 'Learning to Adventure', summary: 'Your first adventure.', author: 'Todd Medema', publishedurl: 'quests/learning_to_adventure.xml', minplayers: 1, maxplayers: 6, mintimeminutes: 20, maxtimeminutes: 30, genre: 'Drama', contentrating: 'Everyone', language: 'English' },
   {id: '0B7K9abSH1xEOWVpEV1JGWDFtWmc', partition: 'expedition-public', theme: 'horror', official: true, title: 'Learning 2: The Horror', summary: 'Your first adventure continues with Expedition: The Horror.', author: 'Todd Medema', publishedurl: 'quests/learning_to_adventure_2_the_horror.xml', expansionhorror: true, minplayers: 1, maxplayers: 6, mintimeminutes: 20, maxtimeminutes: 40, genre: 'Drama', contentrating: 'Everyone', language: 'English' },
+  // TODO make actual intro Future quest
```
smartin015: Should we dev-gate this?

services/app/src/components/base/Dialogs.tsx
```diff
@@ -111,7 +111,10 @@ export class ExpansionSelectDialog extends React.Component<ExpansionSelectDialog
           <Button className="primary large" onClick={() => this.props.onExpansionSelect({horror: false})}>Expedition</Button>
           <br/>
           <br/>
-          <Button className="primary large" onClick={() => this.props.onExpansionSelect({horror: true})}><span className="line">Expedition&nbsp;</span><span className="line">+ The Horror</span></Button>
+          <Button className="primary large" onClick={() => this.props.onExpansionSelect({horror: true})}>Expedition + <img className="inline_icon" src="images/horror_small.svg"/></Button>
```
smartin015: I'm not sure if icons are the most understandable thing to see when a new user opens the app for the first time... what does this look like rendered?

services/app/src/components/views/QuestPreview.test.tsx
```diff
@@ -63,6 +64,9 @@ describe('QuestPreview', () => {
     expect(wrapper.html()).toContain('book');
   });
 
+  it('indicates that horror and future expansions are required', () => {
+    // TODO
```
smartin015: implement?

services/quests/src/components/DialogsContainer.tsx
```diff
@@ -19,7 +19,11 @@ const mapStateToProps = (state: AppState): DialogsStateProps => {
 const mapDispatchToProps = (dispatch: Redux.Dispatch<any>): DialogsDispatchProps => {
   return {
     handleMetadataChange: (quest: QuestType, key: string, value: any): void => {
+      console.log();
```
smartin015: rm

toddmedema:
@smartin015 tests update:
1. added a stub since to track we don't have any implemented test examples for quest creator dialogs & the whole metadata system will be changing soon when we move off Google Docs
2. the app UI currently does not support searching for "Future but not Horror" quests. Since Horror cards are required for Future cards, all Future quests also require Horror
3. Looked into it and this is effectively covered by existing stubs and tests, but added a few more stubs just to be sure

Here's the screenshots. Doing it with words would definitely be clearer, but was also struggling to get the words to appear in a nice way on mobile. Another approach would be to switch off buttons and to checkboxes to select each expansion.
screen shot 2018-08-31 at 19 48 59
screen shot 2018-08-31 at 19 49 05

services/api/src/models/Quests.test.ts
toddmedema: Done

services/api/src/models/Quests.ts
toddmedema: Was trying to be clever with the whole Future requiring Horror thing, but esp since this is an admin, simplified it to just say Horror y/n, Future y/n

services/app/src/Constants.tsx
toddmedema: Good idea, done

services/app/src/components/base/Dialogs.tsx
toddmedema: (see screenshot and idea in main comment)

services/quests/src/components/DialogsContainer.tsx
toddmedema: Done

services/app/src/components/views/QuestPreview.test.tsx
toddmedema: Done

smartin015:
What if instead of buttons, used a picker here? Since there's only a couple choices and a bit of a progression, it might be a better fit and we could also add some detail text for each selection (e.g. just show "Future" as the choice, but in the description say something to the effect of "base game plus horror plus future".

Side note: WDYT about using the synth icon instead of tech as "the icon" for the Future? The wrench is too close to a settings icon IMO.

toddmedema:
I don't think picker is the UI component you meant? https://material-ui.com/demos/pickers/

Ah yep, updated to synth

smartin015:
Picker: https://github.com/ExpeditionRPG/expedition/blob/master/services/app/src/components/base/Picker.tsx

smartin015:
It's a little unfortunate to have a UI where users are allowed to make incompatible choices that we only tell them about after the fact...

I think if we go with checkboxes, we should add a bit of logic to auto-check dependencies and note that those are dependencies of the particular expansion being checked (e.g. click Future, it auto-checks and disables Horror with a message somewhere on the screen indicating why that happened.

toddmedema:
That's fair... kept thinking about this, and here's my proposed solution:

Since we still only have three options, keeping it to the buttons but using the expansion names instead of icons (replacing "Expedition" with "Base" and removing "The"s so they fit on mobile). For this few options, this provides maximum clarity and one-click selection. Once we add another expansion, we can revisit the UI ideas in this thread depending on if it's an optional booster or another in the required chain?

screen shot 2018-09-03 at 10 02 21

PR #439, opened 2018-09-01
Migrate app to Quest schema, improve search results page when no results or when errors.

smartin015:
image

Also includes an updated & shared search interface between the API server and some code to hide the requirements section of the quest details when there's no additional requirements.

Unit tests stubbed, implementation to follow in a separate PR as it's easier to split the `Search` component up and remove the switch-based rendering to keep unit testing light.

### DIFF
--- a/services/app/src/actions/Search.tsx
+++ b/services/app/src/actions/Search.tsx
@@ -1,7 +1,8 @@
+import {QuestSearchResponse} from 'api/Handlers';
 import * as Redux from 'redux';
+import {Quest} from 'shared/schema/Quests';
 import {openSnackbar} from '../actions/Snackbar';
 import {AUTH_SETTINGS, FEATURED_QUESTS} from '../Constants';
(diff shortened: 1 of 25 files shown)


services/app/src/actions/Search.tsx
```diff
@@ -74,22 +73,29 @@ function getSearchResults(params: SearchSettings, callback: (quests: QuestDetail
...
     xhr.setRequestHeader('Content-Type', 'text/plain');
     xhr.onload = () => {
-      const response: any = JSON.parse(xhr.responseText);
-      if (response.error) {
-        dispatch({type: 'SEARCH_ERROR'});
-        return dispatch(openSnackbar(Error('Search error: ' + response.error)));
+      let response: QuestSearchResponse;
+      try {
+        response = JSON.parse(xhr.responseText);
+        if (response.error) {
```
toddmedema: If catch sets response.error, should we check for response.error after the try / catch?

PR #442, opened 2018-09-03
add single space line support via \n

toddmedema:


services/quests/docs/qdl_guide.md
```diff
@@ -501,6 +503,17 @@ _Another card_ (#another)
 After clicking "Jump me!", the app will immediately proceed to this card.
 ```
 
+## Formatting
```
smartin015: Nice. Thoughts on adding something like this to the template/starter quest when you first load the quest creator?

PR #452, opened 2018-09-10
Add tests for decision card template

smartin015:
This adds another 21 tests, bringing our stub (i.e. critical tests) coverage from 51.5% to 53.4%.


services/app/src/components/views/quest/cardtemplates/decision/Actions.test.tsx
```diff
@@ -1,39 +1,212 @@
...
+import {ParserNode} from '../TemplateTypes';
+import {EMPTY_DECISION_STATE} from './Types';
+import {Action, newMockStore} from 'app/Testing';
+import {initialMultiplayer} from 'app/reducers/Multiplayer';
+import {Outcome} from 'shared/schema/templates/Decision';
+
+const cheerio = require('cheerio');
+const seedrandom = require('seedrandom');
+
+const TEST_SETTINGS = {
```
toddmedema: I've noticed that we seem to duplicate TEST_SETTINGS in a lot of places (which makes it less robust against adding / changing how settings work). Thoughts re: exporting this from the settings reducer?

services/app/src/components/views/quest/cardtemplates/decision/Actions.test.tsx
```diff
@@ -1,39 +1,212 @@
...
+    test('returns 3 semi-unique, generated checks', () => {
+      expect(generateLeveledChecks(5, seedrandom.alea('1234'))).toEqual([
+        {difficulty: 'hard', persona: 'light', requiredSuccesses: 1, skill: 'athletics'},
+        {difficulty: 'medium', persona: 'dark', requiredSuccesses: 3, skill: 'athletics'},
+        {difficulty: 'easy', persona: 'light', requiredSuccesses: 1, skill: 'knowledge'},
+      ]);
+    });
+    test('scales num successes with num adventurers', () => {
+      expect(Math.max(...generateLeveledChecks(1, seedrandom.alea('1234')).map((c) => c.requiredSuccesses))).toEqual(1);
+      expect(Math.max(...generateLeveledChecks(6, seedrandom.alea('1234')).map((c) => c.requiredSuccesses))).toBeGreaterThan(3);
```
toddmedema: Wait, so with 6 players.... at least 4 players would have to roll and succeed? That seems a bit much, I would expect 3 at most, both statistically and so that it doesn't slow down the game too much with all the extra rolling

services/app/src/components/views/quest/cardtemplates/decision/Actions.test.tsx
```diff
@@ -1,39 +1,212 @@
...
+    });
     test.skip('scales difficulty with the number of times this type of check was selected previously', () => { /* TODO */ });
   });
   describe('skillTimeMillis', () => {
-    test.skip('respects settings', () => { /* TODO */ });
-    test.skip('scales with player count', () => { /* TODO */ });
+    test('scales with player count', () => {
+      expect(skillTimeMillis(TEST_SETTINGS, TEST_RP)).toBeLessThan(skillTimeMillis({...TEST_SETTINGS, numPlayers: 1}, initialMultiplayer));
```
toddmedema: TEST_SETTINGS has 3 players.... so you have less time with more players? That's the opposite of scaling with player count, isn't it?

If anything, I would expect this to work the same as the combat timer, where 1 player gets 2x the time as everyone else, but the time doesn't change with the # of players (thereby giving smaller parties a slight communication advantage to compensate for their disadvantage in numbers)

services/app/src/components/views/quest/cardtemplates/decision/Actions.test.tsx
smartin015: I'm not a fan of including test-only code in the same file as business logic... makes it more likely to be accidentally used in the wrong way. Added a common `TestData` file instead for the reducers directory, which is what we do in the api code when we need to test a bunch of different kinds of models.

services/app/src/components/views/quest/cardtemplates/decision/Actions.test.tsx
smartin015: Capped at 3.

services/app/src/components/views/quest/cardtemplates/decision/Actions.test.tsx
smartin015: You've actually described exactly the test behavior... refactored to make it more clear.

smartin015:
Updated, PTAL

PR #461, opened 2018-09-15
Refactor & Test Audio Component

smartin015:
This is a major refactor of `<Audio/>`.

- Splits the audio component into smaller testable classes (`AudioNode`, `ThemeManager`, and `Audio` component)
  - Adds a battery of tests for each of these parts.
- Limits scope of existing code to only music (no unused sfx code, shorter variable names).
- Moves audio loading code into `actions/Audio.tsx` to reduce complexity of audio component code
  - Audio is stored in the form of `AudioNode` instances and a `ThemeManager` instance in a new redux store parameter (`audioData`) that is not captured by history logic.


smartin015:
FYI we're now at 55.8% test stub coverage, up from 53.4%.

PR #465, opened 2018-09-16
skill check help text, improved Resolve card

smartin015:
image

image

Also fixes the extant icon coloring issues on the resolution page.


services/app/src/components/views/quest/cardtemplates/decision/PrepareDecision.tsx
```diff
@@ -25,12 +25,22 @@ export default function prepareDecision(props: Props): JSX.Element {
...
     helpText = (
       <div>
+        <h2>
+          Skill Check
+        </h2>
+        <p>
+          You must choose between three actions, then roll to see if your choice succeeds.
+        </p>
+        <p>
+          Each action has a skill type. You get bonuses if your party's skills align with the action you choose,
```
toddmedema: You get bonuses for your levels in skills of those types, and for matching the persona type.

Issue #467, opened 2018-09-16
Bug with card creator: printing zoomed out?


toddmedema:
Have to set the zoom to 150% when printing in Chrome (and we don't support other browsers) - used to work fine...

PR #473, opened 2018-09-22
fix card creator zoom bug

toddmedema:
Deployed to the beta card creator, give it a try: http://betacards.expeditiongame.com

Closes https://github.com/ExpeditionRPG/expedition/issues/467


services/cards/src/components/TopBar.tsx
```diff
@@ -61,26 +62,30 @@ class TopBar extends React.Component<Props, {}> {
...
-            <IconButton onClick={this.props.openHelp}>
-              <HelpOutline />
-            </IconButton>
-          </Tooltip>
-        </Toolbar>
-      </AppBar>
-    );
+    if (this.props.printing) {
+      return null;
+    } else {
```
smartin015: This doesn't need to be in an `else` bracket if the initial conditional returns early.

services/cards/src/reducers/CombinedReducers.tsx
```diff
@@ -1,8 +1,10 @@
 import { combineReducers } from 'redux';
 import Cards from './Cards';
 import Filters from './Filters';
+import Ui from './Ui';
```
smartin015: I don't think `Ui.tsx` was committed.... I'd also recommend changing this to e.g. `Layout` as "Ui" seems overly general.

PR #475, opened 2018-09-22
deploy script arguments

toddmedema:


scripts/deploy.sh
```diff
@@ -1,28 +1,53 @@
...
-        echo "Prod build cancelled until tested on beta."
-      fi
+TARGETS="beta prod local-beta local-prod"
+service="$1"
+target="$2"
+
+getService() {
+  if [ ! -d "$service" ]
+  then
+    echo "What service would you like to deploy?"
```
smartin015: I'd suggest a bracketed list afterwards for clarity (e.g. `[app, quests, ...`)

scripts/deploy.sh
```diff
@@ -1,28 +1,53 @@
...
+      service=$s
+      break
+    done
+  fi
+}
+
+getTarget() {
+  if [[ ! $TARGETS =~ $target ]];
+  then
+    echo "Where would you like to deploy $service?"
```
smartin015: Ditto

toddmedema:
From IRL, script generates and displays options as part of the select command

Issue #342, opened 2018-07-03
Highlight "New" quests 


toddmedema:
Would be a good way to indicate to users that more content is always coming out. 

Thinking we could use the user's last login date, and return the # of public, publishes quests created since then as part of the login API call (or the follow up /user/quests call, since /login is also used by the quest creator)

The tricky thing is that we now sort quests by highest rated first - but, now that all quests have ratings and we show the ratings info even if a quest only has one rating, it might be more ok to switch back to newest-first as the default sort. Then we could either show a "NEW" badge on quests with created time < lastLogin time or, since the quest result cards are getting a bit crowded, show it as a "section", ie a `<hr>NEW</hr>` above the first new quest, and a plain `<hr/>` after the last new quest.

PR #478, opened 2018-09-23
highlight new quests

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition/issues/342

Note that it uses created instead of published (since published is actually "last updated") - this will be more conservative and may occasionally result in a false negative, ie if an author accidentally publishes then unpublishes a quest early in writing, and then re-publishes it months later, anyone who's logged in in the last few months won't see it as "new". The alternative would be to show "NEW" any time it's been updated more recently than the most recent login, but that seems like it would be uselessly noisy. 

screen shot 2018-09-23 at 14 20 04
screen shot 2018-09-23 at 14 19 54

### DIFF
--- a/services/api/src/models/Quests.ts
+++ b/services/api/src/models/Quests.ts
@@ -95,8 +95,8 @@ export function searchQuests(db: Database, userId: string, params: QuestSearchPa
   const order = [];
   if (params.order) {
     if (params.order === '+ratingavg') {
-      // Default search - also show very new quests on top
-      order.push(Sequelize.literal(`NOW() - created <= interval '24 hours' DESC`));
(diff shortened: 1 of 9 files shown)


services/api/src/models/Quests.ts
```diff
@@ -95,6 +95,8 @@ export function searchQuests(db: Database, userId: string, params: QuestSearchPa
   const order = [];
   if (params.order) {
     if (params.order === '+ratingavg') {
+      // Default search - also show very new quests on top
+      order.push(Sequelize.literal(`NOW() - created <= interval '24 hours' DESC`));
```
smartin015: I don't s'pose I could get a unit tests for this new behavior?

Issue #477, opened 2018-09-23
replace Realtime API + Drive with Postgres storage


toddmedema:
## Stage 1: Run duplicate Postgres storage in background

- [x] Save to realtime, drive and postgres, but don't change loading code.
- [x] Save to Postgres in Markdown - two text columns (qdl + notes)
  - [x] Double buffered save (2 rows, saves write the oldest row if both older than 24h, otherwise most recent row)

Key Result: Monitor for a while to make sure infra and code is stable, and that DB storage use isn't utterly insane (which it shouldn't be, since a single quest should only take one row / overwrite it)

## Stage 2 (Dec 11): Switch to loading contents from Postgres by default, but still open files from Drive

- [x] Fall back to realtime, then to loading from drive file if postgres load fails
- [x] Keep front-end code structured as realtime change manager to make it easier to re-add realtime later

## Stage 3 (Dec 11): Turn off realtime

- [x] get rid of writing to realtime (so we'll need to educate about new filesystem + loss of realtime collaboration)
- [x] Add start-of-edit timestamp to questdata when loaded, prevent user with different timestamp from editing the file and indicate in the UI (require a refresh).

## Notes

- we'll still have to support loading files from Drive + content from Realtime API for the foreseeable future

PR #479, opened 2018-09-23
Add quest data to postgres, initial Realtime API Editable replacement.

smartin015:
Editable class is not yet finished, but figured I'd check it in early for later development.

This adds a `questdata` table that saves the user's QDL (and notes). Double buffering discussed in #477 is used when quest is updated. This is stage 1 of migrating off the realtime API.


services/api/src/models/QuestData.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      .then((results) => {
+        expect(results).toContainEqual(jasmine.objectContaining({created: qd.basic.created}));
+        expect(results).toContainEqual(jasmine.objectContaining({created: old2.created}));
+        expect(results).not.toContainEqual(jasmine.objectContaining({created: new1.created}));
+        done();
+      })
+      .catch(done.fail);
+    });
+
+    test('does not affect other users\' saved quests', (done) => {
```
toddmedema: More broadly, does not affect other saved quests, even if owned by the same user

smartin015: Added a test specifically for same user.

services/quests/src/Editable.tsx
```diff
@@ -0,0 +1,78 @@
...
+  constructor(editables: Array<Editable<any>>) {
+    for (const e of editables) {
+      e.setModelHook((v: any) => this.onSetValue(e.name, v));
+    }
+  }
+
+  public onSetValue(k: string, v: any) {
+    const newModel = {...this.history[this.history.length - 1], [k]: v};
+    this.history.push(newModel);
+    // TODO enforce max history size
```
toddmedema: Can we add max model size enforcement here so that we don't forget about it? Even something excessive like 500 could prevent users from experiencing inexplicable pain :)

smartin015: As discussed - this code isn't actually even used yet. Will follow up when implemented.

PR #484, opened 2018-09-25
fix unpublished bug

toddmedema:
closes https://github.com/ExpeditionRPG/expedition/issues/483


toddmedema:
Apparently we already have a stub, lol - `test.skip('removes a set tombstone', () => { /* TODO */ });`

Issue #450, opened 2018-09-10
Move Private Quests to top of search results, label with "PRIVATE"


toddmedema:
Should also add a checkbox filter, enabled by default, for "Include my private quests" to provide clarity and customization. 

Also check the quest creator help docs for any references to private publishing and update them to this new system. We'll also want to make sure to highlight this in a Quest Crafter email shortly after it's released!

PR #490, opened 2018-10-05
Added checkbox for filtering private quests

Saumya29:
### [Issue #450](https://github.com/expeditionRPG/expedition/issues/450) 

* **What kind of change does this PR introduce?**
Feature request

* **What is the current behavior?**
Earlier, there wasn’t an option for the user to configure the settings for viewing private/public quests.

* **What I did?**
Added a checkbox for the user to enable Private quests and displayed them on the top.
This checkbox is added on the "Filter & Sort" page ie SearchSettings component!

### DIFF
--- a/services/api/src/Handlers.ts
+++ b/services/api/src/Handlers.ts
@@ -5,7 +5,7 @@ import * as Joi from 'joi';
 import * as memoize from 'memoizee';
 import * as request from 'request-promise';
 import {AnalyticsEvent} from 'shared/schema/AnalyticsEvents';
-import {PRIVATE_PARTITION, PUBLIC_PARTITION} from 'shared/schema/Constants';
+import {PUBLIC_PARTITION} from 'shared/schema/Constants';
 import {Feedback} from 'shared/schema/Feedback';
 import {QuestData} from 'shared/schema/QuestData';
 import {Quest} from 'shared/schema/Quests';
@@ -157,25 +157,19 @@ export function search(db: Database, req: express.Request, res: express.Response
     players: body.players,
     requirespenpaper: body.requirespenpaper,
     text: body.text,
+    showPrivate: body.showPrivate,
   };
 
-  if (params.partition === PUBLIC_PARTITION && res.locals.id) {
-    // Also search for private quests by this user
-    const privateParams = {...params, owner: res.locals.id, partition: PRIVATE_PARTITION};
-    return Promise.all([
-      doSearch(db, res.locals.id, params),
-      doSearch(db, res.locals.id, privateParams),
-    ]).then((results) => {
-      res.status((results[0].error) ? 500 : 200).end(JSON.stringify({
-        error: results[0].error,
-        hasMore: results[0].hasMore,
-        quests: [...results[1].quests, ...results[0].quests],
-      }));
-    });
-  }
-
   return doSearch(db, res.locals.id, params).then((result) => {
-    res.status((result.error) ? 500 : 200).end(JSON.stringify(result));
+    const sortedQuests: Quest[] = result.quests.sort((a: Quest, b: Quest) => {
+      if (a.partition.toLowerCase() <= b.partition.toLowerCase()) {
+        return -1;
+      } else {
+        return 1;
+      }
+    });
+    const sortedResult: QuestSearchResponse = {...result, quests: sortedQuests};
+    res.status((result.error) ? 500 : 200).end(JSON.stringify(sortedResult));
   });
 }
 
--- a/services/api/src/models/Quests.test.ts
+++ b/services/api/src/models/Quests.test.ts
@@ -1,24 +1,22 @@
-import {PUBLIC_PARTITION} from 'shared/schema/Constants';
-import {Quest} from 'shared/schema/Quests';
-import {QuestInstance} from './Database';
-import {searchQuests} from './Quests';
-import {
-  quests as q,
-  testingDBWithState,
-} from './TestData';
+import { object } from 'joi';
+import { PRIVATE_PARTITION, PUBLIC_PARTITION } from 'shared/schema/Constants';
+import { Quest } from 'shared/schema/Quests';
+import { QuestInstance } from './Database';
+import { searchQuests } from './Quests';
+import { quests as q, testingDBWithState } from './TestData';
 
 const Moment = require('moment');
 
 describe('quest', () => {
   describe('searchQuests', () => {
-    const quests = [q.basic, q.horror, q.future];
+    const quests = [q.basic, q.private, q.horror, q.future];
 
     test('returns an empty array if no results', (done: DoneFn) => {
       testingDBWithState(quests)
-        .then((tdb) => {
-          return searchQuests(tdb, '', {partition: 'otherpartition'});
+        .then(tdb => {
+          return searchQuests(tdb, '', { partition: 'otherpartition' });
         })
-        .then((results) => {
+        .then(results => {
           expect(results.length).toEqual(0);
           done();
         })
@@ -27,8 +25,8 @@ describe('quest', () => {
 
     test('returns full quest data', (done: DoneFn) => {
       testingDBWithState(quests)
-        .then((tdb) => {
-          return searchQuests(tdb, '', {partition: PUBLIC_PARTITION});
+        .then(tdb => {
+          return searchQuests(tdb, '', { partition: PUBLIC_PARTITION });
         })
         .then((results: QuestInstance[]) => {
           expect(results.length).toEqual(1);
@@ -43,12 +41,14 @@ describe('quest', () => {
 
     test('does not return expansions if unspecified', (done: DoneFn) => {
       testingDBWithState(quests)
-        .then((tdb) => {
-          return searchQuests(tdb, '', {partition: PUBLIC_PARTITION});
+        .then(tdb => {
+          return searchQuests(tdb, '', { partition: PUBLIC_PARTITION });
         })
         .then((results: QuestInstance[]) => {
           expect(results.length).toEqual(1);
-          expect((results[0] as any).dataValues).toEqual(jasmine.objectContaining({id: 'questid'}));
+          expect((results[0] as any).dataValues).toEqual(
+            jasmine.objectContaining({ id: 'questid' }),
+          );
           done();
         })
         .catch(done.fail);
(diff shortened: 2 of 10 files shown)


Saumya29:
ezgif com-video-to-gif

services/app/src/actions/Search.tsx
```diff
@@ -63,48 +64,62 @@ export const searchAndPlay = remoteify(function searchAndPlay(id: string, dispat
...
-      }
+    // TODO: Create a type for questSearchPromises
+    const questSearchPromises: any = [];
 
-      // Log silently to console, so we can see if users file feedback
-      // with recurrent search problems.
-      if (response.error) {
-        console.error(response.error);
-      }
+    const partitionQuestsPromise = fetch(AUTH_SETTINGS.URL_BASE + '/quests', {
```
toddmedema: I like this improvement, `fetch` is much better than our old `xhr` function

services/app/src/actions/Search.tsx
```diff
@@ -63,48 +64,62 @@ export const searchAndPlay = remoteify(function searchAndPlay(id: string, dispat
...
+      method: 'POST',
+      headers: {
+        'Content-Type': 'text/plain',
+      },
+      body: JSON.stringify(params),
+    });
+
+    questSearchPromises.push(partitionQuestsPromise);
+
+    if (params.partition === 'expedition-private') {
```
toddmedema: Rather than having this occur as two separate requests, I was thinking we could pass a new search parameter that says "return results across both partitions" and keep it as a single request. Something like `showprivate`

services/app/src/components/views/Search.tsx
```diff
@@ -202,6 +210,8 @@ export class SearchSettingsCard extends React.Component<SearchSettingsCardProps,
               <option value="false">No</option>
             </NativeSelect>
           </FormControl>
+          <Checkbox label="Enable Private Quests" value={this.state.partition === 'expedition-private' ? true : false} onChange={(v: boolean) => { this.handleCheckbox(v); }}>
```
toddmedema: Should be "Also show my private quests"

services/app/src/actions/Search.tsx
Saumya29: On it

services/api/src/Handlers.ts
```diff
@@ -155,25 +155,21 @@ export function search(db: Database, req: express.Request, res: express.Response
...
-      }));
-    });
-  }
-
   return doSearch(db, res.locals.id, params).then((result) => {
-    res.status((result.error) ? 500 : 200).end(JSON.stringify(result));
+    const sortedQuests: Quest[] = result.quests.sort((a: Quest, b: Quest) => {
+      const questAPartition = a.partition.toLowerCase();
+      const questBPartition = b.partition.toLowerCase();
+      if (questAPartition <= questBPartition) {
```
toddmedema: This can be streamlined to `if (a.partition.toLowerCase() <= b.partition.toLowerCase()) {`; it's only a few characters longer but saves two lines, and those variables aren't re-used :)

services/api/src/models/Quests.ts
```diff
@@ -41,7 +42,9 @@ export function searchQuests(db: Database, userId: string, params: QuestSearchPa
   // TODO: Validate search params
   const where: Sequelize.WhereOptions<Partial<Quest>> = {published: {$ne: null} as any, tombstone: null};
 
-  where.partition = params.partition || PUBLIC_PARTITION;
+  if (params.showPrivate === false) {
+    where.partition = params.partition || PUBLIC_PARTITION;
```
toddmedema: If showPrivate is true, we should explicitly only search for quests where partition is PUBLIC_PARTITION or PRIVATE_PARTITION, since we may add future partitions in the future. Link for the Sequelize `Op.or` syntax: http://docs.sequelizejs.com/manual/tutorial/querying.html

We should also make sure to sort by partition first, so that private quests are listed first, if showPrivate is true.

services/app/src/actions/Search.tsx
```diff
@@ -67,43 +68,35 @@ function getSearchResults(params: SearchParams, callback: (response: QuestSearch
...
-        for (const q of response.quests) {
+
+      if (data.quests) {
+        for (const q of data.quests) {
           const i = Quest.create(q);
           if (i instanceof Error) {
             console.error('Parsed invalid quest: ' + JSON.stringify(q));
-          } else {
-            quests.push(i);
+           } else {
```
toddmedema: looks like some extra whitespace accidentally got inserted here?

Saumya29: Yes, will take care of it.

services/api/src/models/Quests.ts
Saumya29: On it.

services/api/src/models/Quests.test.ts
```diff
@@ -11,14 +11,15 @@ const Moment = require('moment');
...
-    const quests = [q.basic, q.horror, q.future];
+    const quests = [q.basic, q.private, q.horror, q.future];
 
     test('returns an empty array if no results', (done: DoneFn) => {
       testingDBWithState(quests)
         .then((tdb) => {
           return searchQuests(tdb, '', {partition: 'otherpartition'});
         })
         .then((results) => {
+          console.log(results);
```
toddmedema: remove this console log since it was just for debugging :)

services/api/src/models/Quests.test.ts
```diff
@@ -65,6 +66,49 @@ describe('quest', () => {
         .catch(done.fail);
     });
 
+    test('return private quests (alongwith public quests) when showPrivate is set to true', (done: DoneFn) => {
+      testingDBWithState(quests)
+        .then((tdb) => {
+          return searchQuests(tdb, '', {partition: PUBLIC_PARTITION, showPrivate: true});
+        })
+        .then((results) => {
+          console.log(results);
```
toddmedema: Another console log we can remove

services/api/src/models/Quests.test.ts
```diff
@@ -65,6 +66,49 @@ describe('quest', () => {
...
+          expect(results.length).toEqual(2);
```
toddmedema: Note that this test could use a more robust `expect`, since the length check will fail every time a new quest is added to this section. Instead, we should confirm that at least one result has partition = public, and at least one = private

services/api/src/models/Quests.test.ts
```diff
@@ -65,6 +66,49 @@ describe('quest', () => {
...
+    });
+
+    test('doesn\'t return private quests when showPrivate is set to false', (done: DoneFn) => {
+      testingDBWithState(quests)
+        .then((tdb) => {
+          return searchQuests(tdb, '', {partition: PUBLIC_PARTITION, showPrivate: false});
+        })
+        .then((results) => {
+          expect(results.length).toEqual(1);
+          expect(results[0].partition).toEqual(PUBLIC_PARTITION);
```
toddmedema: Similar thought here; rather than checking for length and partition of one quest, we can make this more robust by just confirming that all results have partition = public

services/api/src/models/Quests.ts
```diff
@@ -41,7 +44,17 @@ export function searchQuests(db: Database, userId: string, params: QuestSearchPa
   // TODO: Validate search params
   const where: Sequelize.WhereOptions<Partial<Quest>> = {published: {$ne: null} as any, tombstone: null};
 
-  where.partition = params.partition || PUBLIC_PARTITION;
+  if (params.showPrivate === true) {
+    where.partition = {
+      [Op.or]: [PUBLIC_PARTITION, PRIVATE_PARTITION],
+    };
+
+    // where.order = [
```
toddmedema: Please remove commented out code :)

services/api/src/models/Quests.ts
```diff
@@ -41,7 +44,17 @@ export function searchQuests(db: Database, userId: string, params: QuestSearchPa
...
+    //   ['partition', 'ASC'],
+    // ];
+  } else {
+    where.partition =  params.partition || PUBLIC_PARTITION;
```
toddmedema: extra space after the equals

services/api/src/models/TestData.ts
```diff
@@ -77,8 +77,39 @@ const basicQuest = new Quest({
   userid: 'testuser',
 });
 
+const basicPrivateQuest = new Quest({
```
toddmedema: Recommend changing this to how we did the horror quest below, where it extends the base quest. Less code to update when we add a new field to the quest object!

services/app/src/actions/Search.tsx
toddmedema: Hmm, still seeing the extra whitespace here?

services/api/src/models/Quests.test.ts
Saumya29: Yes, makes sense

services/api/src/models/Quests.ts
Saumya29: Sorry about that 😅

services/api/src/models/Quests.ts
toddmedema: That is awesome! Definitely going to do that :D

toddmedema:
@Saumya29 ping on this one, would be awesome to get it wrapped up :)

Saumya29:
@toddmedema Can you review this?

services/api/src/models/Quests.test.ts
```diff
@@ -1,34 +1,35 @@
...
-import {Quest} from 'shared/schema/Quests';
-import {QuestInstance} from './Database';
-import {searchQuests} from './Quests';
 import {
-  quests as q,
-  testingDBWithState,
-} from './TestData';
+  PUBLIC_PARTITION,
+  PRIVATE_PARTITION
+} from "../../../../shared/schema/Constants";
```
toddmedema: We've generally styled things to use single quotes instead of double quotes, looks like you might need to update your editor / linter?

services/api/src/models/Quests.test.ts
toddmedema: Looks like this still needs to be updated to be more generic - iterate through all results and expect each result's partition to equal public

services/api/src/models/Quests.test.ts
```diff
@@ -41,98 +42,240 @@ describe('quest', () => {
...
+        })
+        .then(results => {
+          expect(results.length).toEqual(2);
+          Object.keys(results[0]).forEach(key => {
+            if(results[0][key].hasOwnProperty('partition')) {
+              expect(results[0][key].partition).toEqual(PRIVATE_PARTITION);
+            }
+          })
+          Object.keys(results[1]).forEach(key => {
+            if(results[1][key].hasOwnProperty('partition')) {
```
toddmedema: code style: space after if: `if (results)`

services/app/src/components/views/Search.test.tsx
```diff
@@ -1,12 +1,12 @@
 import * as React from 'react';
-import {LanguageType} from 'shared/schema/Constants';
-import {Quest} from 'shared/schema/Quests';
+import {LanguageType} from '../../../../../shared/schema/Constants';
```
toddmedema: what's going on here with all of the `../`'s? We should support the `shared/` alias shortcut

Saumya29:
@toddmedema Please review

services/api/src/models/Quests.test.ts
toddmedema: Looks like this still needs to be fixed?

services/app/src/actions/Search.tsx
toddmedema: Now that we have prettier set up, can you run it through this file again to sort out the extraneous spaces?

Issue #457, opened 2018-09-15
Improve performance on search API route


toddmedema:
Appears to be taking 300ms+ just for the API server to respond.

Probably just needs an index on the DB. If that doesn't fix it, there might be something slow about how Heroku does their DB setup? Worth trying what the shortest latency on a simple request (like get 1 element from a 1 element table) is

PR #494, opened 2018-10-06
UI V2

### LINKED ISSUE #451: Implement bottom tab navigation (UI 2.0)
Tabs along the bottom for Tutorials, Search, Saves (+Offline), History, Game Masters / GM's Corner

https://material-ui.com/demos/bottom-navigation/

Ideally also supports swiping between the views.

As part of this, I suppose we can move the Shop link to the top right menu. GM Mode will go to the Dungeon Masters panel, and private quests will become part of search results: https://github.com/ExpeditionRPG/expedition/issues/450

We should save the current panel to local storage each time it changes, and use that value on app load and any other time we need a default value for which panel to go to.

smartin015:
Still TODO:
- [ ] Add actual GM corner quests
- [x] See what breaks in remote play
- [x] Test all the things

Upgrades:
- bottom nav bar with tutorial, offline, quest search, history, and GM corner buttons
- new `<QuestButton/>` component that unifies view of quests
- new custom icons for new quests, official quests, completed quests, offline quests, gm quests, quest history, and private quests
- (refactor) split search component into individual components
- extended `<Card/>` to allow setting `null` for `onReturn` to disable return button
- Added a BTREE index to the quests table, reducing lookup latency of the default search from 330ms to 160ms
- Includes the user's private quests at the front of the search results.

Fixes #451 
Closes https://github.com/ExpeditionRPG/expedition/issues/457


services/api/src/Handlers.ts
```diff
@@ -156,6 +156,22 @@ export function search(db: Database, req: express.Request, res: express.Response
...
+    // Also search for private quests by this user
+    const privateParams = {...params, owner: res.locals.id, partition: PRIVATE_PARTITION};
+    return Promise.all([
+      doSearch(db, res.locals.id, params),
+      doSearch(db, res.locals.id, privateParams),
+    ]).then((results) => {
+      res.status((results[0].error) ? 500 : 200).end(JSON.stringify({
+        error: results[0].error,
+        hasMore: results[0].hasMore,
+        quests: [...results[1].quests, ...results[0].quests],
```
toddmedema: Is this guaranteed to return private then public?

toddmedema:
Also, looks like Saved & Offline still has a back button in the top left (whereas the rest don't). Without the back button, it's also unclear how to get back to the player select screen... Perhaps we could replace the back button on all those screens with a home icon in the top left, that just takes you back to the opening screen?

services/api/src/Handlers.ts
smartin015: yep

smartin015:
Updated to remember the last tab - tested behavior on double-tap start, tap & hold start, multiplayer start, and return from quest.

PR #499, opened 2018-10-14
add multiplayer / doubletap note to splash screen

toddmedema:
screen shot 2018-10-14 at 10 41 24
screen shot 2018-10-14 at 10 41 30

### DIFF
--- a/services/cards/src/helpers.tsx
+++ b/services/cards/src/helpers.tsx
@@ -87,8 +87,8 @@ export function translateTier(tier: number, english: string, translations: Trans
 // TODO modernize
 export function healthCounter(health: number, back = false): JSX.Element {
 
-  const digitWidth = [0, 16, 23];
-  let maxWidth = 269;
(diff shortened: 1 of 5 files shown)


services/cards/src/helpers.tsx
```diff
@@ -87,8 +87,8 @@ export function translateTier(tier: number, english: string, translations: Trans
 // TODO modernize
 export function healthCounter(health: number, back = false): JSX.Element {
 
-  const digitWidth = [0, 16, 23];
-  let maxWidth = 269;
```
smartin015: what's up with these card creator changes?

toddmedema: Woops, thought I'd put them elsewhere. Fixes two bugs where the 0/30 didn't line up on encounter backs, and a few HP amounts (if I recall, 24 and 26) had weird overlaps on the bottom track. Tested by making a sheet with enemies with HP 0->30 and they all looked good.

PR #500, opened 2018-10-14
search: fix bug with re-searching, preserve language

toddmedema:
Bug was that going back from settings was not trigger a new search.

Streamlined the search code to use an onMount trigger rather than trying to catch every single place where a goToSearch is called; that, along with clearing the search state each time a setting or search param is changed should make it much more robust (ie now you can go home, change player count and go back, and it will definitely re-search).

Also now stores and uses search language; decided not to also restore other search params, since those are more likely to change / break, and also more likely to change across search requests.

screen shot 2018-10-14 at 12 00 13

### DIFF
--- a/services/app/src/actions/Search.tsx
+++ b/services/app/src/actions/Search.tsx
@@ -4,13 +4,17 @@ import {Quest} from 'shared/schema/Quests';
 import {openSnackbar} from '../actions/Snackbar';
 import {AUTH_SETTINGS, TUTORIAL_QUESTS} from '../Constants';
 import {ExpansionsType, SearchParams, SettingsType} from '../reducers/StateTypes';
-import {SearchResponseAction} from './ActionTypes';
-import {remoteify} from './ActionTypes';
+import {remoteify, SearchChangeParamsAction, SearchResponseAction} from './ActionTypes';
+import {} from './ActionTypes';
 import {toCard} from './Card';
 import {previewQuest} from './Quest';
 
+export function changeSearchParams(params: any): SearchChangeParamsAction {
+  return {type: 'SEARCH_CHANGE_PARAMS', params};
+}
+
 // TODO: Make search options propagate to other clients
-export const search = remoteify(function search(a: {params: SearchParams, settings: SettingsType}, dispatch: Redux.Dispatch<any>) {
+export const search = remoteify(function searchAndView(a: {params: SearchParams, settings: SettingsType}, dispatch: Redux.Dispatch<any>) {
   const params = {...a.params};
   Object.keys(params).forEach((key: string) => {
     if ((params as any)[key] === null) {
@@ -19,7 +23,6 @@ export const search = remoteify(function search(a: {params: SearchParams, settin
   });
   params.players = a.settings.numPlayers;
   params.expansions = Object.keys(a.settings.contentSets).filter( (key) => a.settings.contentSets[key] ) as ExpansionsType[],
-  dispatch(toCard({name: 'SEARCH_CARD'}));
   dispatch(getSearchResults(params, (response: QuestSearchResponse) => {
     dispatch({
       quests: response.quests,
--- a/services/app/src/components/views/Search.tsx
+++ b/services/app/src/components/views/Search.tsx
@@ -14,11 +14,12 @@ export interface StateProps {
   params: SearchParams;
   settings: SettingsType;
   user: UserState;
-  results: Quest[];
+  results: Quest[]|null;
   searching: boolean;
 }
 
 export interface DispatchProps {
+  search: (params: SearchParams, settings: SettingsType) => void;
   toCard: (name: CardName) => void;
   onReturn: () => void;
   onQuest: (quest: Quest) => void;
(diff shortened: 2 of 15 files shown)


services/app/src/actions/Search.tsx
```diff
@@ -4,13 +4,17 @@ import {Quest} from 'shared/schema/Quests';
...
 import {toCard} from './Card';
 import {previewQuest} from './Quest';
 
+export function changeSearchParams(params: any): SearchChangeParamsAction {
+  return {type: 'SEARCH_CHANGE_PARAMS', params};
+}
+
 // TODO: Make search options propagate to other clients
-export const search = remoteify(function search(a: {params: SearchParams, settings: SettingsType}, dispatch: Redux.Dispatch<any>) {
+export const search = remoteify(function searchAndView(a: {params: SearchParams, settings: SettingsType}, dispatch: Redux.Dispatch<any>) {
```
smartin015: searchAndView, but you've removed toCard from this?

services/app/src/components/views/Search.tsx
```diff
@@ -98,43 +99,72 @@ function renderNoQuests(props: Props): JSX.Element {
...
-export function search(props: Props): JSX.Element {
-  if (!props.user.loggedIn) {
-    return renderStoredQuests(props);
-  }
+export class Search extends React.Component<Props, {}> {
 
-  if (props.searching) {
-    return renderLoading(props);
+  public componentDidMount() {
+    if (!this.props.results) {
```
smartin015: AIUI this trips when a user isn't logged in (where we should just be showing stored quests). Does this cause an error? Can we test this?

services/app/src/reducers/Search.tsx
```diff
@@ -1,11 +1,13 @@
...
+import {SearchChangeParamsAction, SearchResponseAction} from '../actions/ActionTypes';
+import {getStorageString, setStorageKeyValue} from '../LocalStorage';
 import {SearchState} from './StateTypes';
 
 export const initialSearch: SearchState = {
-  results: [],
+  results: null, // null = need to search; [] = no results
   params: {
-    language: 'English',
+    language: getStorageString('search-language', 'English') as LanguageType,
```
smartin015: We should probably put `search-language` in CONSTANTS.

services/app/src/reducers/Search.tsx
```diff
@@ -14,11 +16,21 @@ export const initialSearch: SearchState = {
...
 export function search(state: SearchState = initialSearch, action: Redux.Action): SearchState {
   switch (action.type) {
+    case 'CHANGE_SETTINGS':
+      // Clear results when invalidated.
+      return {...state, results: null};
+    case 'SEARCH_CHANGE_PARAMS':
+      // Update params and clear results
+      const changes = (action as SearchChangeParamsAction).params || {};
+      Object.keys(changes).forEach((key: string) => {
+        setStorageKeyValue(`search-${key}`, changes[key]);
```
smartin015: Seems like we're only using `search-language`... I'd be more comfortable if this just set that, until/unless we need the others for something.

services/app/src/components/views/Search.tsx
toddmedema: Good thought; just tested, nothing breaks in incognito

smartin015: I mean a unit test... thoughts:
- doesn't trigger a search attempt when user not logged in
- searches when no search results on initial page view
- doesn't search when there are already search results

services/app/src/reducers/Search.tsx
```diff
@@ -1,11 +1,13 @@
...
+import {SearchChangeParamsAction, SearchResponseAction} from '../actions/ActionTypes';
+import {getStorageString, setStorageKeyValue} from '../LocalStorage';
 import {SearchState} from './StateTypes';
 
 export const initialSearch: SearchState = {
-  results: [],
+  results: null, // null = need to search; [] = no results
   params: {
-    language: 'English',
+    language: getStorageString('language', 'English') as LanguageType,
```
smartin015: This should still be a file-level constant value IMO.

toddmedema: Should we change how do it for settings too then? https://github.com/ExpeditionRPG/expedition/blob/master/services/app/src/reducers/Settings.tsx#L9

smartin015: Probably a good idea, yeah. Seems like we should have a `defaultSettings` const with the defaults, and then an `initialSettings` which is a map of the `defaultSettings` keys, resolving via local storage where possible.

toddmedema: From IRL, defaultSettings stuff would be a lot of work, so maybe later

toddmedema:
Updated, PTAL

services/app/src/components/views/Search.test.tsx
```diff
@@ -39,6 +41,22 @@ describe('Search', () => {
     return {props, e: render(<Search {...(props as any as Props)} />, undefined /*renderOptions*/)};
   }
 
+  function setupMount(overrides?: Partial<Props>) {
```
smartin015: This is a lot of duplicate code IMO... easy for it to drift (for example, onSearch only exists in `setupMount`).

I'd prefer if you changed the original `setup()` to `mount` the component, then where we need to render things just do `setup().render()` or equivalent.

services/app/src/components/views/Search.test.tsx
```diff
@@ -56,4 +74,39 @@ describe('Search', () => {
...
+    const {props, e} = setupMount({
+      user: {
+        loggedIn: true,
+      },
+      results: null,
+    });
+    setTimeout(() => {
+      expect(props.onSearch).toHaveBeenCalled();
+      done();
+    }, 100);
```
smartin015: Fighting timeouts with timeouts is a pretty huge anti-pattern in tests (easy to have race conditions).

What if instead we moved the async `setTimeout` logic to the container, so we can easily detect whether onSearch is called synchronously or not?

services/app/src/components/views/Search.tsx
```diff
@@ -102,11 +102,12 @@ function renderNoQuests(props: Props): JSX.Element {
 export class Search extends React.Component<Props, {}> {
 
   public componentDidMount() {
+    console.log('mount', this.props.results, this.props.user);
```
smartin015: rm

services/app/src/reducers/Search.tsx
```diff
@@ -4,10 +4,12 @@ import {SearchChangeParamsAction, SearchResponseAction} from '../actions/ActionT
 import {getStorageString, setStorageKeyValue} from '../LocalStorage';
 import {SearchState} from './StateTypes';
 
+const LANGUAGE = 'language';
```
smartin015: maybe `LANGUAGE_KEY`? Otherwise it seems like the language itself is a constant.

services/app/src/reducers/Search.tsx
```diff
@@ -4,10 +4,12 @@ import {SearchChangeParamsAction, SearchResponseAction} from '../actions/ActionT
...
+
 export const initialSearch: SearchState = {
   results: null, // null = need to search; [] = no results
   params: {
-    language: getStorageString('language', 'English') as LanguageType,
```
smartin015: What parses `English`? should we be concerned about matching case here and elsewhere? Can we pull this default from a constant somewhere else?

toddmedema: The casting itself doesn't actually guarantee it'll be correct, true. That being said, 'English' is used 21 times in the codebase, and I'm sure the problem is in several other places, so I've filed a code cleanliness issue to address it more holistically: https://github.com/ExpeditionRPG/expedition/issues/502

PR #504, opened 2018-10-16
Force sync button for multiplayer

smartin015:
- Added a sync button so users can manually sync state
- Fix multi_event handling to actually work (was giving events in descending order which the app was rejecting)
- commitID and snackbar state now persist across RETURN.
- Fixed sync ui (spinner now shows fraction of events replayed, and actually shows up)
- Small renames (inFlight -> commitID, INFLIGHT_* -> MULTIPLAYER_*
- Pulled out event handler and status code from multiplayer client, put into a react component for testing
- Add tests for MULTI_EVENT, ACTION, and reducer RETURN code

image
image


toddmedema:
(once tests passing, of course)

PR #512, opened 2018-10-27
add Card Creator support for combining spreadsheets

toddmedema:
Needed for generating production print files for deluxe edition. Confirmed that it loads single and multiple sheets, and non-English sheets.


services/cards/src/actions/Cards.tsx
```diff
@@ -8,10 +8,51 @@ import {filtersCalculate} from './Filters';
...
+  cards: CardType[];
+  translations: TranslationsType;
+}
+
 export function downloadCards(): ((dispatch: Redux.Dispatch<any>) => void) {
   return (dispatch: Redux.Dispatch<any>) => {
     const store = getStore();
     dispatch(cardsLoading());
+
+    const keys = store.getState().filters.source.current.split(':')[1].split(',');
```
smartin015: Any chance this could be passed into the action?

services/cards/src/actions/Cards.tsx
```diff
@@ -8,10 +8,51 @@ import {filtersCalculate} from './Filters';
...
+
+    Promise.all(keys.map(downloadAndProcessSpreadsheet))
+      .then((results: any[]) => {
```
smartin015: This is a rather complex type to leave as `any`. Could you make this an interface?

services/cards/src/actions/Cards.tsx
```diff
@@ -8,10 +8,51 @@ import {filtersCalculate} from './Filters';
...
+
+        const translations = results.reduce((acc: TranslationsType, obj: ResultType) => {
+          return {...acc, ...obj.translations};
+        }, {});
+
+        if (Object.keys(translations).length > 0) {
+          dispatch(translationsUpdate(translations));
+        }
+        dispatch(cardsUpdate(cards));
+        dispatch(cardsFilter(store.getState().cards.data, store.getState().filters));
```
smartin015: This is also a lot of unnecessary `getState` for an action. Can we pass these in as arguments?

toddmedema: So, this is how it was done previously, since cardsUpdate() actually causes data mutations that must be read into cardsFilter. Agree that it's not the greatest, but redoing the card processing / filtering system is out of scope here :(

services/cards/src/actions/Cards.tsx
toddmedema: We could, but that would duplicate this logic into three places... which seems like it would cause its own set of problems

toddmedema:
Commented and updated

services/cards/src/actions/Cards.tsx
smartin015: I was thinking just pass the filters into `downloadCards`, not the keys themselves. The logic doesn't need to be duplicated that way.

services/cards/src/actions/Cards.tsx
smartin015: Woah, did not realize that -  it's not indicated by this code in any way and I don't think we have any other redux action that intentionally rely on their own side effects like this. This is a regression waiting to happen, since we also have no implemented unit tests for this code whatsoever.

If you look at what `cardsUpdate` does... it just sets `state.cards.data` to `cards` and doesn't even touch `state.filters`. So we really don't need to re-fetch the args for `cardsFilter` at all. Can you fix that?

`filtersCalculate()` is a bit trickier, so I'm OK with a big fat TODO comment indicating why we're getting state like this... provided we actually unit test this function and open a bug to pull the logic out of the reducer and into an action function.

services/cards/src/actions/Cards.tsx
```diff
@@ -37,23 +79,19 @@ export function downloadCards(): ((dispatch: Redux.Dispatch<any>) => void) {
...
         // Note: doesn't yet have access to sheet name, that's assigned in callback
         return card;
       },
       simpleSheet: true,
     });
-  };
+  });
 }
 
 export interface CardsLoadingAction extends Redux.Action {
```
smartin015: Mind quickly moving these to an `ActionTypes` file to be consistent with all our other services?

PR #515, opened 2018-10-29
Fix hot module reloading

### LINKED ISSUE #460: Fix hot module reloading
When you edit a component, webpack should be able to properly reload just that component, not force refresh the entire app. Will make development much faster!

Saumya29:
Closes #460 

giphy


Saumya29:
**Cards**

cards

**Quests**

questsvideo

Issue #234, opened 2018-07-01
Fix multiplayer edge cases


toddmedema:
<a href="https://github.com/smartin015"><img src="https://avatars2.githubusercontent.com/u/607666?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [smartin015](https://github.com/smartin015)**
_Sunday Mar 18, 2018 at 22:45 GMT_
_Originally opened as https://github.com/ExpeditionRPG/expedition-app/issues/615 (1 comment(s))_

----

- [x] exiting multiplayer does not actual exit 
- [x] Disable multiplayer button when in multiplayer state
- [x] Fix button IDs when different fundamental app settings  (hashed key for button IDs, such as quest title).
- [x] Compute shared settings for multiplayer (e.g. intersection of app expansion list) and display it on the lobby page.
- [x] shouldn't be able to go to "select adventurers" page during remote play (after splash screen double tap)
- [x] Clear history when starting multiplayer (prevents back button weirdness)
- [x] Prevent multiplayer if app is outdated
- [x] Adventurer count & tier sum pickers are buggy (they jump around when tapped and don't actually change values)
- [x] Multiplayer exit does not reset stats counters
- [x] Skill check timer page didn't show same values across MP session
- [x] In bad wifi, multiplayer connection was not shown to be broken

Probably not needed:
- Search should allow changing # players without affecting the app player count
- Search page doesn't show horror + future even though both clients were set to it. We should have a dropdown for manually selecting some combo of expansions
- Server should recognize when both clients are on same event ID, but different quest line numbers (and do what? replay past 5 actions? force resync?)

toddmedema:
<a href="https://github.com/toddmedema"><img src="https://avatars2.githubusercontent.com/u/775657?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [toddmedema](https://github.com/toddmedema)**
_Saturday Apr 07, 2018 at 19:12 GMT_

----

Added that exiting remote play does not actually exit

PR #522, opened 2018-11-10
Refactor & Test Multiplayer

smartin015:
Big PR, LMK if you want this chunked out a bit.

Changes:
- Moved `remotify` code from ActionTypes.tsx to multiplayer/Remoteify.tsx
- Added a bunch of tests
- Action-ified `handleEvent`, `sendEvent`, `sendStatus` in `actions/Multiplayer.tsx` (instead of having them be client methods)
- Passed default client to actions so they can also be injected during tests
- Moved multiplayer counters from client to `multilplayer/Counters.tsx`
- Tried to remove occurrences of `getMultiplayerConnection` (that aren't in `actions/Multiplayer.tsx`) where possible
  - This made `MultiplayerAffector` require a container so it can properly dispatch actions
- Greatly simplified multiplayer status code and multiplayer client code
- Changed multiplayer state to include client and instance ids, connection state

#234

### DIFF
--- a/services/app/package.json
+++ b/services/app/package.json
@@ -32,9 +32,5 @@
   "dependencies": {
     "cordova-android": "^7.1.1",
     "cordova-ios": "^4.5.5"
-  },
-  "devDependencies": {
-    "webpack-cli": "^3.1.2",
-    "webpack-command": "^0.4.1"
   }
 }


services/app/package.json
```diff
@@ -32,5 +32,9 @@
   "dependencies": {
     "cordova-android": "^7.1.1",
     "cordova-ios": "^4.5.5"
+  },
+  "devDependencies": {
+    "webpack-cli": "^3.1.2",
```
toddmedema: Shouldn't these be in the root package.json?

services/app/src/components/multiplayer/MultiplayerClientContainer.tsx
```diff
@@ -1,36 +1,34 @@
...
   return {
-    onMultiEventStart: (syncID: number) => {
-      dispatch(local({type: 'MULTIPLAYER_MULTI_EVENT_START', syncID} as MultiplayerMultiEventStartAction));
+    onStatus: (client?: string, instance?: string, status?: StatusEvent) => {
+      dispatch(sendStatus(client, instance, status));
     },
-    onMultiEventComplete: () => {
-      dispatch(local({type: 'MULTIPLAYER_MULTI_EVENT'}));
+    onEvent: (e: MultiplayerEvent, buffered: boolean, commitID: number, multiplayer: MultiplayerState): Promise<void> => {
+      return dispatch(handleEvent(e, buffered, commitID, multiplayer));
```
toddmedema: Don't need to return the dispatch here, right?

smartin015: Correct - also found a bug in status which I fixed and tested.

PR #525, opened 2018-11-11
Added clarification on quest goto errors

### LINKED ISSUE #336: Small: Make 'goes to nowhere' error more helpful
_Sunday Apr 29, 2018 at 19:45 GMT_
_Originally opened as https://github.com/ExpeditionRPG/expedition-quest-creator/issues/483 (0 comment(s))_

----

Would save user frustration and employee time debugging if the "goes to nowhere" error could specifically shout out what #ID doesn't exist if it's a GOTO at fault.

I feel like I've also seen at least one case where that check didn't work (perhaps the playtester ran out of time before it got to the problem node?), so it might be worth switching this check up to be a simplified scan for `**goto (.*)**` and a matching `#($1)` in the raw text

smartin015:
image

Fixes #336

### DIFF
--- a/shared/render/render/XMLRenderer.tsx
+++ b/shared/render/render/XMLRenderer.tsx
@@ -153,7 +153,7 @@ export const XMLRenderer: Renderer = {
         return;
       }
       if (rendered.find('#' + m[1]).length === 0) {
-        log.err('goto "' + m[1] + '" without matching ID (check your spelling)', '426', parseInt(c.attribs['data-line'], 10) || 0);
+        log.err('goto "' + m[1] + '" does not match any card IDs (check your spelling)', '426', parseInt(c.attribs['data-line'], 10) || 0);
       }
     });
     return [];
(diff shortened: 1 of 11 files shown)


shared/render/render/XMLRenderer.tsx
```diff
@@ -132,4 +133,29 @@ export const XMLRenderer: Renderer = {
...
+
+    // Ensure no incorrectly named GOTOs
+    rendered.find('trigger').each((i, c) => {
+      const text = cheerio(c).text();
+      const m = text.match(/goto (.*)/);
+      if (m === null) {
+        return;
+      }
+      if (rendered.find('#' + m[1]).length === 0) {
+        log.err('goto "' + m[1] + '" without matching ID (check your spelling)', '426', parseInt(c.attribs['data-line'], 10) || 0);
```
toddmedema: `goto bad does not match any card IDs (check your spelling)`

PR #529, opened 2018-11-18
Filter out reviews with no extra details

smartin015:
To stop the tide of useless rating emails...


toddmedema:
Might be nice to add a comment to the code since this is a bit magical

PR #537, opened 2018-11-20
fix: update the api readme and config setupdb

### LINKED ISSUE #536: Steps for setup local env under API service are depreciated

nimish-gupta:
Fixes #536

### DIFF
--- a/services/api/src/config.ts
+++ b/services/api/src/config.ts
@@ -10,6 +10,7 @@ nconf.argv()
   // Quest storage config settings
   'DATABASE_URL', // URL of postgres database storing quest metadata, user data, feedback, etc.
   'SEQUELIZE_LOGGING', // Enable console logging of sequelize SQL queries
+  'SEQUELIZE_SSL',
 
   // Feedback email sender config settings
   'MAIL_EMAIL',
@@ -64,6 +65,7 @@ nconf.argv()
   PORT: 8081,
   SEQUELIZE_LOGGING: true,
   SUPER_USER_IDS: [],
+  SEQUELIZE_SSL: true,
 });
 
 export default nconf;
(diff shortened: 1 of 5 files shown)


services/api/src/config.ts
```diff
@@ -10,6 +10,7 @@ nconf.argv()
   // Quest storage config settings
   'DATABASE_URL', // URL of postgres database storing quest metadata, user data, feedback, etc.
   'SEQUELIZE_LOGGING', // Enable console logging of sequelize SQL queries
+  'SEQUELIZE_SSL',
```
nimish-gupta: For local setup, secure SSL setup is not required.

services/api/src/config.ts
```diff
@@ -64,6 +65,7 @@ nconf.argv()
   PORT: 8081,
   SEQUELIZE_LOGGING: true,
   SUPER_USER_IDS: [],
+  SEQUELIZE_SSL: true,
```
nimish-gupta: Setup default to true, so that we don't have to do any changes in the production

services/api/src/models/Database.ts
```diff
@@ -77,13 +77,15 @@ export class Database {
       ],
       timestamps: false, // TODO: eventually switch to sequelize timestamps
     });
+    this.analyticsEvent.sync();
```
nimish-gupta: Added these because on starting the app, tables were not being created.

services/api/src/lib/oauth2.ts
```diff
@@ -77,10 +78,11 @@ export function installOAuthRoutes(db: Database, router: express.Router) {
   // then they will be redirected to that URL when the flow is finished.
   // This also fetches their info from the DB and returns that (if available).
   // [START authorize]
-  router.post('/auth/google', // LOGIN
+  router.post('/auth/google', limitCors, // LOGIN
```
nimish-gupta: As it was throwing error `No access-control-allow-origin` error on local

PR #538, opened 2018-11-20
feat: added the quest filter

### LINKED ISSUE #403: Quest search filter: Uses The Horror Expansion
Yesterday I was looking for a quest that specificially does use the horror expansion, would be nice to be able to search for it. We'd probably want to set this up in such a way that it can be used for other expansions in the future.

nimish-gupta:
Fixes #403


nimish-gupta:
@toddmedema right now only `horror` expansion is hard-coded https://github.com/ExpeditionRPG/expedition/blob/master/services/app/src/reducers/StateTypes.tsx#L88-L91 in the `contentSets`.
So for now, to populate the list of expansions in the search filter, I have hard-coded `future` and `horror` expansions.

nimish-gupta:
Also, most of the formatting related changes are done using `prettier vscode extension`.

andrewlinfoot:
@nimish-gupta you should either: disable the extension, do a PR setting up Prettier to their ESLint rules or do a PR and run it on the whole repo. This way you don't have some files that use one style-guide and a different style for the rest (prettier vs non-prettier)

nimish-gupta:
> @nimish-gupta you should either: disable the extension, do a PR setting up Prettier to their ESLint rules or do a PR and run it on the whole repo. This way you don't have some files that use one style-guide and a different style for the rest (prettier vs non-prettier)

@Saumya29 is working on setuping prettier. From this PR, I will just remove the prettier changes.

toddmedema:
Sounds good

services/app/src/components/views/ExpansionCheckbox.tsx
```diff
@@ -0,0 +1,108 @@
...
+  onChange: (values: string[]) => void;
+  value: string[] | undefined;
+}
+
+export class ExpansionCheckbox extends React.Component<IProps, {}> {
+  public state: IState;
+  constructor(props: IProps) {
+    super(props);
+    const isHorror = this.props.contentSets.horror ? true : false;
+    const isFuture = this.props.contentSets.future ? true : false;
```
toddmedema: Shouldn't you be able to use content sets.future directly as a Boolean?

nimish-gupta: So by default `future` is undefined.

nimish-gupta:
I am getting build https://www.dropbox.com/s/d5dd6dm4t8c3mao/Screenshot%202018-11-21%2002.30.48.png?dl=0 under Travis. Am I missing any dependency? (Because it's working locally)

toddmedema:
Ah. So. Looking at the loom, we actually don't want to use tooltips since our userbase is almost entirely mobile. Much better to include the "enable in settings" text directly into the disabled checkboxes' labels.

nimish-gupta:
> Ah. So. Looking at the loom, we actually don't want to use tooltips since our userbase is almost entirely mobile. Much better to include the "enable in settings" text directly into the disabled checkboxes' labels.

Ohh got it, probably this will also resolve the travis error. Thanks :)

package.json
```diff
@@ -23,8 +23,8 @@
     "quests-local": "API_HOST='http://localhost:8081' npm run quests & npm run api"
   },
   "dependencies": {
-    "@material-ui/core": "^1.2.1",
-    "@material-ui/icons": "^1.1.0",
+    "@material-ui/core": "^3.5.1",
```
toddmedema: With such a large version increase in such a core package, can you run through our beta test checklist to make sure the rest of the app + quest creator UI still looks and behaves reasonably? https://docs.google.com/document/d/1HkHct0-AEMaHpkpU0uqSi6mhj2GJId78bXc23mo-2qg/edit

nimish-gupta: Can you please grant access to `nimish@pesto.tech` for the above doc?

Also, I have bumped this version just for figuring out the travis issue with tooltip. Probably I will change this to before as we are not requiring tooltip.

services/app/src/components/views/ExpansionCheckbox.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+    const props: IProps = {
+      onChange: jest.fn(),
+      contentSets: { horror: true },
+      value: undefined,
+    };
+    const elem = mount(<ExpansionCheckBox {...props} />);
+    return {elem, props};
+  }
+
+  test('changes search param values when onChange is clicked', () => {
```
toddmedema: Can you also add a test to confirm the desired initial behavior based on settings? (i.e. that if the content set is enabled, it starts checked, and if it's disabled, it mentions settings in the text)

nimish-gupta:
https://www.useloom.com/share/3635e25edd7440928032d9141eaa6ddc
in the above video, I have put console.log just before the query is sent to the server for quests. So it's working as per desired behaviour

nimish-gupta:
@toddmedema I don't have merging permissions.

PR #540, opened 2018-11-21
Ask users for feedback when exiting a quest early

### LINKED ISSUE #498: Ask whether a quest was too hard, too broken, etc. when user returns home out of a quest
In order to describe what they think of a quest, users have to complete the quest. especially for broken quests, we should allow users to tell us what they thought about it even if they don't complete it and instead head back to the home screen.

smartin015:
Fixes #498

### DIFF
--- a/services/app/src/components/base/DialogsContainer.tsx
+++ b/services/app/src/components/base/DialogsContainer.tsx
@@ -27,7 +27,17 @@ const mapStateToProps = (state: AppState): StateProps => {
   };
 };
 
-const mapDispatchToProps = (dispatch: Redux.Dispatch<any>): DispatchProps => {
+function validateFeedback(text: string): Error|null {
+  if (!text) {
+    return new Error('Please enter a description so that we can help resolve the issue.');
+  }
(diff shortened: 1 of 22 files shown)


services/app/src/components/base/DialogsContainer.tsx
```diff
@@ -27,7 +27,17 @@ const mapStateToProps = (state: AppState): StateProps => {
   };
 };
 
-const mapDispatchToProps = (dispatch: Redux.Dispatch<any>): DispatchProps => {
+function validateFeedback(text: string): Error|null {
+  if (!text) {
```
toddmedema: We shouldn't block people exiting a quest if they don't want to type something - for example, someone could just be quickly trying out the quest to see if they like it

smartin015: We don't - this is used in `onFeedbackSubmit`, not in `onExitQuest`. I double-checked in a local instance that you don't need to give feedback to exit the quest. Also updated "submit" button to "exit" in the dialog so it makes more sense.

PR #546, opened 2018-11-22
Quest data: prevent concurrent edits

### LINKED ISSUE #477: replace Realtime API + Drive with Postgres storage

smartin015:
Fixes #477

As part of this PR I had to allow the API server to respond to OPTIONS requests with access control headers - this allows us to get back nonzero status codes when sending CORS requests to the API server.

image


services/quests/src/actions/Quest.tsx
```diff
@@ -170,8 +170,9 @@ function createDocNotes(model: any) {
   return str;
 }
 
-function loadQuestFromAPI(user: UserState, docid: string): Promise<{data: string, notes: string, metadata: any}|null> {
-  return fetch(API_HOST + '/qdl/' + docid, {
+function loadQuestFromAPI(user: UserState, docid: string, edittime: Date): Promise<{data: string, notes: string, metadata: any, edittime: Date}|null> {
+  console.log(edittime.getTime());
```
toddmedema: Rm the extra console logs here and elsewhere

smartin015: Done.

smartin015:
It manifests as a "save error", meaning the editor state is kept dirty. Added a screenshot to the top comment and confirmed that the "changes not saved" alert prompt works as intended.

PR #547, opened 2018-11-23
feat: added the accounts page

### LINKED ISSUE #384: App: User Account page
Currently I don't think there's a way to do any of the following in the app. I think all of this could be built into an "account" page:

- See what account you are logged in as
- Log out
- See how many loot points you have
- View recently played quests / view quest history + feedback and ratings submitted

nimish-gupta:
Fixes #384

### DIFF
--- a/services/api/src/models/Users.ts
+++ b/services/api/src/models/Users.ts
@@ -1,5 +1,5 @@
 import * as Bluebird from 'bluebird';
-import Sequelize from 'sequelize';
+import Sequelize, {Op} from 'sequelize';
 import {PUBLIC_PARTITION} from 'shared/schema/Constants';
 import {Quest} from 'shared/schema/Quests';
 import {User} from 'shared/schema/Users';
@@ -42,18 +42,24 @@ export function subscribeToCreatorsList(mc: any, email: string) {
   });
 }
 
-export interface UserQuestsType {
-  [questId: string]: {
-    details: Quest;
-    lastPlayed: Date;
-  };
+interface IUserQuestType {
+  details: Quest;
+  lastPlayed: Date;
 }
 
-export function getUserQuests(db: Database, id: string): Bluebird<UserQuestsType> {
+export interface UserQuestsType {
+  [questId: string]: IUserQuestType;
+}
+// Added this questId to filter the user posts on the basis of quests on which user had given the feedback.
+export function getUserQuests(db: Database, id: string, questIds?: string[]): Bluebird<UserQuestsType> {
+  const where = {userID: id, category: 'quest', action: 'end', questID: { [Op.notIn]: [] as string[] }};
+  if (questIds) {
+    where.questID = { [Op.in]: questIds };
+  }
   return db.analyticsEvent.findAll({
+    where,
     attributes: ['questID', [Sequelize.fn('MAX', Sequelize.col('created')), 'lastPlayed']],
     group: 'questID',
-    where: {userID: id, category: 'quest', action: 'end'},
   })
   .then((results: any[]) => {
     const userQuests = {} as UserQuestsType;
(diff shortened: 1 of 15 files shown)


services/api/src/models/Users.ts
```diff
@@ -42,18 +42,24 @@ export function subscribeToCreatorsList(mc: any, email: string) {
...
 export interface UserQuestsType {
-  [questId: string]: {
-    details: Quest;
-    lastPlayed: Date;
-  };
+  [questId: string]: IUserQuestType;
 }
 
-export function getUserQuests(db: Database, id: string): Bluebird<UserQuestsType> {
+export function getUserQuests(db: Database, id: string, questIds?: string[]): Bluebird<UserQuestsType> {
```
nimish-gupta: Added this questId to filter the user posts on the bases of quests on which user had given the feedback.

nimish-gupta:
https://www.dropbox.com/s/iim7zed4slvtzs5/Screenshot%202018-11-23%2006.11.46.png?dl=0

services/api/src/models/Users.ts
toddmedema: Would be good to add this as a comment to the code so that we remember why this happened in the future :)

services/app/src/actions/Card.tsx
```diff
@@ -17,6 +17,11 @@ export interface ToCardArgs {
 export const toCard = remoteify(function toCard(a: ToCardArgs, dispatch: Redux.Dispatch<any>, getState?: () => AppStateWithHistory): ToCardArgs {
   const nav = getNavigator();
   const state = (getState !== undefined) ? getState() : getStore().getState();
+  // Disable clicking on the same menu again
```
toddmedema: Messing with the toCard action is pretty dangerous since it affects the functionality of the entire app. What problem were you seeing that made you want to add this?

services/app/src/actions/User.tsx
```diff
@@ -182,10 +184,36 @@ function updateState(dispatch: Redux.Dispatch<any>): ((u: UserState) => Promise<
...
+
+type TReduxThunk<ReturnType> = (dispatch: Redux.Dispatch<any>, getState: () => AppState) => ReturnType;
+
+export function logoutUser(): TReduxThunk<Promise<void>> {
+  return (dispatch) => {
+    const gapi = getGapi();
+    const auth2 = gapi.auth2.getAuthInstance();
+    return auth2.signOut().then(() => {
+      dispatch({type: 'USER_LOGOUT'});
+      dispatch(toCard({name: 'TUTORIAL_QUESTS'}));
```
toddmedema: Seeing this, realizing that it would be good to, after changing cards, also show a snackbar saying "Logged out successfully", so that the user is absolutely certain that they were logged out and not just redirected to another page :)

services/app/src/actions/Card.tsx
nimish-gupta: So before this behaviour was like this https://www.useloom.com/share/ac0e84d5fc194eaf98aab1e34819e6cd

after the change
https://www.useloom.com/share/3e4102b97ef64e65946019ccf12de9fd

toddmedema: Gotcha. An even better solution here would be to disable the menu item for the page the user is currently on - not only would that prevent this issue, it would also give the user a visual indication of which menu option / page they're on

nimish-gupta: I will create an issue for this behaviour, as this feature is beyond the scope of this PR.

services/api/src/models/Users.ts
nimish-gupta: Done

services/app/src/actions/Card.tsx
toddmedema: Sounds good. In the meantime let's revert this change to the to card action. Thanks!

nimish-gupta: Yeah I have already reverted it

services/app/src/components/base/Card.tsx
```diff
@@ -137,9 +141,9 @@ class Card extends React.Component<Props, {}> {
             </Menu>
           </span>
           <div className="title">{this.props.title}</div>
-        </div>
-        {this.props.header && <div className="header">{this.props.header}</div>}
-        <div className="article">
+        </div>;
```
toddmedema: Weird, should there be a semicolon here?

services/app/src/components/base/Card.tsx
```diff
@@ -193,8 +197,8 @@ class Card extends React.Component<Props, {}> {
             </div>
           </div>
           {icon}
-        </div>
-      </div>
+        </div>;
```
toddmedema: Is this semicolon expected?

services/app/src/components/views/Account.tsx
```diff
@@ -0,0 +1,84 @@
...
+            <tbody>
+              <tr><th>Name</th><td>{user.name}</td></tr>
+              <tr><th>Email</th><td>{user.email}</td></tr>
+              <tr><th>Loot Points</th><td>{user.lootPoints || 'No points'}</td></tr>
+            </tbody>
+          </table>
+          <br/>
+          <br/>
+          <Divider/>
+          <h3>Feedbacks</h3>
```
toddmedema: Let's change this to `My Reviews`

services/app/src/components/base/Card.tsx
nimish-gupta: wierd, yeah I will remove these semicolons.

Issue #237, opened 2018-07-01
Multiplayer UI/UX Improvements


toddmedema:
<a href="https://github.com/smartin015"><img src="https://avatars2.githubusercontent.com/u/607666?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [smartin015](https://github.com/smartin015)**
_Sunday Mar 18, 2018 at 22:48 GMT_
_Originally opened as https://github.com/ExpeditionRPG/expedition-app/issues/618 (0 comment(s))_

----

- [x] footer middle button / player icons: show player names (emails?) in session
- [x] include a "force sync" button that replays the client's quest up to the sync point
- [x] better handling of large MULTI_ACTION (freezes app)
- [x] make the report button more discoverable
- [x] rating quests at end is weird - first person to submit triggers end for everyone / destroys / prevents them from reviewing
- [x] At multiplayer connect screen, recommend using some other device to handle audio/video
- [x] Where you can change player count (settings + mode select), when playing multiplayer, also show the total number of multiplayer adventurers and block delta increases that would go over 6 for the party. (requires fixing the lag on adventurer count change showing up locally + other clients)
- [x] animation on syncing is a bit choppy/weird.

PR #548, opened 2018-11-25
Show total (inc. multiplayer) player count when changing player count

smartin015:
playercount

Part of #237 

Also adds a bunch of tests for mode select.

### DIFF
--- a/services/app/src/components/views/Settings.tsx
+++ b/services/app/src/components/views/Settings.tsx
@@ -61,7 +61,7 @@ const Settings = (props: Props): JSX.Element => {
 
       <Picker id="playerCount" label="Adventurers" value={props.settings.numLocalPlayers} onDelta={(i: number) => props.onPlayerDelta(props.settings.numLocalPlayers, i)}>
         {(adventurers > 1) ? 'The number of players.' : <div><strong>Solo play:</strong> Play as two adventurers with double the combat timer.</div>}
-        {props.multiplayer && <div>({adventurers} across all clients)</div>}
+        {props.multiplayer && <div>({adventurers} across all devices)</div>}
       </Picker>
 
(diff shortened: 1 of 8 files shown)


services/app/src/components/views/Settings.tsx
```diff
@@ -46,56 +50,57 @@ const timerText: { [v: string]: any } = [
...
-      <p className="expansionLabel">Currently playing: <strong>Expedition {props.contentSets.horror ? <span> + Horror</span> : null}{props.contentSets.future ? <span> + Future</span> : null}</strong></p>
+      <p className="expansionLabel">Currently playing: <strong>Expedition {props.settings.contentSets.horror ? <span> + Horror</span> : null}{props.settings.contentSets.future ? <span> + Future</span> : null}</strong></p>
 
-      <Picker label="Adventurers" value={props.numLocalPlayers} onDelta={(i: number) => props.onPlayerDelta(props.numLocalPlayers, i)}>
-        {(props.numLocalPlayers > 1) ? 'The number of players.' : <div><strong>Solo play:</strong> Play as two adventurers with double the combat timer.</div>}
+      <Picker id="playerCount" label="Adventurers" value={props.settings.numLocalPlayers} onDelta={(i: number) => props.onPlayerDelta(props.settings.numLocalPlayers, i)}>
+        {(adventurers > 1) ? 'The number of players.' : <div><strong>Solo play:</strong> Play as two adventurers with double the combat timer.</div>}
+        {props.multiplayer && <div>({adventurers} across all clients)</div>}
```
toddmedema: Instead of all "clients" would it make sense to refer to them as "devices"? Since clients and devices mean the same thing / have a 1:1 map, and most folks are more familiar with the term / concept of devices?

smartin015: Renamed to devices.

PR #552, opened 2018-11-25
Better multiplayer bug reporting

smartin015:
Remove feedback sending from MP connection icon in favor of regular Submit Feedback flow - also send MP counters whenever they exist as part of feedback.

Part of #237

### DIFF
--- a/services/app/src/actions/Web.tsx
+++ b/services/app/src/actions/Web.tsx
@@ -8,7 +8,7 @@ import {MIN_FEEDBACK_LENGTH} from '../Constants';
 import {getDevicePlatform, getPlatformDump} from '../Globals';
 import {getLogBuffer} from '../Logging';
 import {logEvent} from '../Logging';
-import {getCounters, MultiplayerCounters} from '../multiplayer/Counters';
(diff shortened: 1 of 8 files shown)


services/app/src/actions/Web.tsx
```diff
@@ -147,7 +147,11 @@ export function subscribe(a: {email: string}) {
 }
 
 export function submitUserFeedback(a: {quest: QuestState, settings: SettingsType, user: UserState, type: FeedbackType, anonymous: boolean, text: string, rating: number|null}) {
-  return (dispatch: Redux.Dispatch<any>) => {
+  return (dispatch: Redux.Dispatch<any>, getState: () => AppState) => {
+    const stats = getCounters();
+    if (stats) {
+      logMultiplayerStats(a.user, a.quest.details, stats);
```
toddmedema: To clarify, it looks like these stats are getting logged separate from the users' feedback? Would it be possible to instead inject the info into their feedback, so that it's all in one place?

smartin015: Done.

PR #556, opened 2018-11-25
Show content set intersection on multiplayer lobby page

smartin015:
image

Part of #234

### DIFF
--- a/services/app/src/Constants.tsx
+++ b/services/app/src/Constants.tsx
@@ -192,6 +192,12 @@ export const INIT_DELAY = {
   SILENT_LOGIN_MILLIS: 1000,
 };
 
+export const CONTENT_SET_FULL_NAMES: {[k: string]: string} = {
+  base: 'Expedition',
+  horror: 'The Horror',
+  future: 'The Future',
+};
+
 export const CARD_TRANSITION_ANIMATION_MS = 300;
 export const VIBRATION_SHORT_MS = 30; // for navigation / card changes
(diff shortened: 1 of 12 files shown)


services/app/src/Constants.tsx
```diff
@@ -192,6 +192,12 @@ export const INIT_DELAY = {
   SILENT_LOGIN_MILLIS: 1000,
 };
 
+export const CONTENT_SET_FULL_NAMES: {[k: string]: string} = {
```
toddmedema: Since you're introducing this as a new constant (thank you), could you do a quick find and replace across the code?

services/app/src/components/views/Multiplayer.tsx
```diff
@@ -73,12 +74,42 @@ class MultiplayerConnect extends React.Component<Props, {}> {
...
+export function renderLobby(props: Props): JSX.Element {
   return (
     <Card title="Lobby">
       <div className="remoteplay">
-        <div><strong>Session created!</strong> Tell your friends to connect with the following code:</div>
+        <h2>Multiplayer Session</h2>
+        <div>Tell your friends to connect with the following code:</div>
         <h1 className="sessionCode">{props.multiplayer.session && props.multiplayer.session.secret}</h1>
+        <h2>Content Sets</h2>
+        <p>These are the content sets that every device has:</p>
```
toddmedema: has, and thus will be enabled in the session

services/app/src/Constants.tsx
smartin015: done.

PR #565, opened 2018-11-28
fix: disable the menu item after selection

### LINKED ISSUE #562: Disable menu item while on corresponding menu page.

nimish-gupta:
Fixes #562


services/app/src/components/base/Card.tsx
```diff
@@ -131,8 +141,8 @@ class Card extends React.Component<Props, {}> {
               onClose={() => this.handleMenuClose()}>
               <MenuItem id="homeButton" onClick={() => {this.onMenuSelect('HOME'); }}>Home</MenuItem>
               {this.props.inQuest && <MenuItem onClick={() => {this.onMenuSelect('SAVE'); }}>Save quest</MenuItem>}
```
toddmedema: Out of curiosity, why not also apply this to the other menu items?

And, to confirm, when does this un-disable? For example, if you open settings, then change them and go back, I would expect settings to no longer be disabled.

nimish-gupta: So other menu items were not changing the cards, those were either URLs or dialogs, so i didn't add them.

These links will be disable when the cards are changed, so here linking is between menu items and card title.

Issue #561, opened 2018-11-26
Sentry error: Maximum update depth exceeded


toddmedema:
Looks like hundreds of sessions + users have crashed in the past few weeks due to this bug :(

Init.tsx: https://sentry.io/fabricate-llc/expedition-app/issues/596335507/?query=is:unresolved
Multiplayer/middleware.tsx: https://sentry.io/fabricate-llc/expedition-app/issues/776501551/?query=is:unresolved
Multiplayer/middleware.tsx: https://sentry.io/fabricate-llc/expedition-app/issues/768955947/?query=is:unresolved
Android bundle: https://sentry.io/fabricate-llc/expedition-app/issues/582242098/?query=is:unresolved
Android bundle: https://sentry.io/fabricate-llc/expedition-app/issues/582242116/?query=is:unresolved

smartin015:
The last time this happened was ~7 days ago on android and doesn't have the extra debugging info needed to determine what's going on. Closing this, let's reopen after the next app deploy if it recurs.

PR #571, opened 2018-12-01
Fix return behavior when just after splash screen

smartin015:
Sometimes users can cause a RETURN action to be dispatched on the splash page by rapidly tapping the flyout menu, resulting in an exception of "Cannot read property 'name' of undefined".

#561


services/app/src/Init.tsx
```diff
@@ -163,6 +163,8 @@ function handleUrlHash() {
   }
 }
 
+const MAX_ERROR_SNACKBAR_RATE_MILLIS = 1000;
```
toddmedema: Feel like a more intuitive name is `MAX_SNACKBAR_ERROR_RATE_MILLIS`

PR #577, opened 2018-12-05
Add crawler check for invalid transition between two combats

### LINKED ISSUE #315: Crawler should use mid-combat roleplay flow
<a href="https://github.com/smartin015"></img></a> **Issue by [smartin015](https://github.com/smartin015)**
_Sunday Oct 08, 2017 at 12:55 GMT_
_Originally opened as https://github.com/ExpeditionRPG/expedition-quest-creator/issues/403 (2 comment(s))_

----

Follow starting node's parents up to combat node, as well as target node.
Compare both combat's data-line attribute to confirm equal / same combat.
Otherwise, error

smartin015:
Fixes #315

Note that this handles both combat cases mentioned in the bug, and the op evaluation errors were addressed in a previous github issue.

Also includes a test to catch accidentally committing `test.only` to the repo, as this was accidentally done in a couple places and we had tests we didn't know were failing.

### DIFF
--- a/services/quests/errors/definitions/427.tsx
+++ b/services/quests/errors/definitions/427.tsx
@@ -1,6 +1,6 @@
 export const NUMBER = 427;
-export const NAME = `Detected an invalid transition from this combat to another combat`;
-export const DESCRIPTION = `Jumping between combat nodes is not supported.`;
+export const NAME = `Invalid transition from this combat to another combat`;
+export const DESCRIPTION = `Jumping between combat cards is not supported.`;
 export const TEST_WITH_CRAWLER = true;
 
 export const INVALID = [
@@ -124,7 +124,7 @@ _combat_
 
 * on round
 
-  Users get here by entering the combat node first.
+  Users get here by entering the combat card first.
 
 * on win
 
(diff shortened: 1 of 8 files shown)


services/quests/errors/definitions/427.tsx
```diff
@@ -0,0 +1,136 @@
+export const NUMBER = 427;
+export const NAME = `Detected an invalid transition from this combat to another combat`;
```
toddmedema: Can be streamlined to `Invalid transition from this...`

services/quests/errors/definitions/427.tsx
```diff
@@ -0,0 +1,136 @@
...
+export const DESCRIPTION = `Jumping between combat nodes is not supported.`;
```
toddmedema: Nodes -> cards

services/quests/errors/definitions/427.tsx
smartin015: Done.

services/quests/errors/definitions/427.tsx
smartin015: Done.

Issue #506, opened 2018-10-17
Multiplayer handling content set mismatch & too many players


toddmedema:
1. transmit content sets as part of initial multiplayer message
2. have a dialog for warning folks when there's a content set mismatch, w/ information on the lowest common denominator and how to change (+ block connecting)
3. once you're in a multiplayer session, disable the change content sets button

This way, we don't have to continuously transmit content sets or re-calculate local vs all content sets - once you're in a session, you know you're all locked to the same content set.

Similarly, for number of players, we should show a dialog that says too many players in the session, with a player count selector so that they can change + continue to join; once in a multiplayer session, all player count selectors (aka that selector, plus the game mode screen + settings screen) should be mindful of the total number of multiplayer players, preventing you from increasing it over 6 + showing a note in the subtext when the mp player count is 6 / session is already full.

smartin015:
This is a more detailed version of part of #234, removing it from that bug

PR #580, opened 2018-12-06
Refactor content sets

smartin015:
The "content sets" part of #506. Content sets are transmitted as part of MP status, so they can be changed in multiplayer and it shows in the appropriate places. 

This comes with an abstraction of contentsets driven both by settings and multiplayer redux state, plumbing it through to the necessary places.

image


toddmedema:
(well, once tests pass)

PR #582, opened 2018-12-08
Add player count to MP lobby, prevent too many players

### LINKED ISSUE #506: Multiplayer handling content set mismatch & too many players

smartin015:
Over size:
image

Correct size:
image

Fixes #506 

Includes tests for `<Picker/>` and `ctx._.numAdventurers()`.

### DIFF
--- a/services/app/src/components/base/PlayerCount.tsx
+++ b/services/app/src/components/base/PlayerCount.tsx
@@ -1,0 +1,31 @@
+import * as React from 'react';
+import {MAX_ADVENTURERS} from '../../Constants';
+import Picker from './Picker';
+
+interface Props extends React.Props<any> {
+  allPlayers: number;
+  localPlayers: number;
+  id?: string;
+  onChange: (localPlayers: number) => any;
+}
+
+export default class PlayerCount extends React.Component<Props, {}> {
+  private onDelta(delta: number) {
+    if (delta > 0 && this.props.allPlayers + delta > MAX_ADVENTURERS) {
+      return;
+    }
+    if (delta < 0 && this.props.localPlayers + delta < 1) {
+      return;
+    }
+    this.props.onChange(this.props.localPlayers + delta);
+  }
+
+  public render() {
+    return (
+      <Picker id={this.props.id} label="Players" value={this.props.localPlayers} onDelta={(i: number) => this.onDelta(i)}>
+        {(this.props.allPlayers > 1) ? 'The number of players.' : <div><strong>Solo play:</strong> Play as two adventurers with double the combat timer.</div>}
+        {(this.props.allPlayers > this.props.localPlayers) && <div>({this.props.allPlayers} across all devices)</div>}
+      </Picker>
+    );
+  }
+}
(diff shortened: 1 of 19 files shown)


services/app/src/components/base/PlayerCount.tsx
```diff
@@ -0,0 +1,31 @@
...
+    if (delta < 0 && this.props.localPlayers + delta < 1) {
+      return;
+    }
+    this.props.onChange(this.props.localPlayers + delta);
+  }
+
+  public render() {
+    return (
+      <Picker id={this.props.id} label="Players" value={this.props.localPlayers} onDelta={(i: number) => this.onDelta(i)}>
+        {(this.props.allPlayers > 1) ? 'The number of players.' : <div><strong>Solo play:</strong> Play as two adventurers with double the combat timer.</div>}
```
toddmedema: Technically not just the combat timer, affects the decision timer too... Not quite sure how to say that concisely tho

services/app/src/components/views/Multiplayer.tsx
```diff
@@ -108,8 +113,12 @@ export function renderLobby(props: Props): JSX.Element {
...
         </table>
         <p>Click the connected adventurers or connection state for more information.</p>
-        <p>Once everyone is connected, click Start:</p>
-        <Button id="start" className="mediumbutton" onClick={() => {props.onStart(); }}>Start</Button>
+        <h2>Adventurers</h2>
+        <p>Max of {MAX_ADVENTURERS} players allowed in multiplayer.</p>
+        <PlayerCount id="playerCount" localPlayers={props.settings.numLocalPlayers} allPlayers={allPlayers} onChange={(localPlayers: number) => props.onPlayerChange(localPlayers)} />
+        <br/>
+        <p>Once everyone is ready, click Start:</p>
+        <Button id="start" className="mediumbutton" disabled={allPlayers > MAX_ADVENTURERS} onClick={() => {props.onStart(); }}>{(allPlayers > MAX_ADVENTURERS) ? `Player count must be < ${MAX_ADVENTURERS}` : 'Start'}</Button>
```
toddmedema: Must be <=

services/app/src/components/base/PlayerCount.tsx
smartin015: Removed "combat".

services/app/src/components/views/Multiplayer.tsx
smartin015: Done.

PR #590, opened 2018-12-09
Spinner on splash screen to indicate touch timeout

### LINKED ISSUE #56: App: player select touch screen, show timer spinner around number
So that players can see how much time is left to place their fingers on the screen (like the current timer mechanism, the timer should reset when fingers are added or removed)

smartin015:
- Fixes #56 
- Also fixes the weird flickering of multitouch buttons due to wrong placement in the component lifecycle
- Also also fixes styling on bottom footer (not shown in gif) where it was overridden with a white background for some reason.

splash_spinner

### DIFF
--- a/services/app/src/components/views/SplashScreen.tsx
+++ b/services/app/src/components/views/SplashScreen.tsx
@@ -1,3 +1,4 @@
+import CircularProgress from '@material-ui/core/CircularProgress';
 import * as React from 'react';
 import {DOUBLE_TAP_MS, SPLASH_SCREEN_TIPS} from '../../Constants';
 import {AnnouncementState} from '../../reducers/StateTypes';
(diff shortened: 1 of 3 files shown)


toddmedema:
Screenshot plz? Also, can you test this on the oldest phone you can find in the house, just to make sure it doesn't break the user experience from a performance perspective? :)

services/app/src/components/views/SplashScreen.tsx
```diff
@@ -48,10 +51,22 @@ class PlayerCounter extends React.Component<PlayerCounterProps, {}> {
...
+      this.animate();
     }
     this.setState({touchCount: numFingers, maxTouches: Math.max(this.state.maxTouches, numFingers)});
   }
 
+  private animate() {
+    if (this.state.transitionTimeout === null) {
+      return;
+    }
+    const progress = (Date.now() - this.state.lastTouchTime) / (this.props.transitionMillis) * 100;
```
toddmedema: If performance is an issue, this would be straigthforward to move to being a CSS animation that gets reset by JS via a class change

smartin015: It slows down a little when there's multiple touches, but the performance is roughly the same as what happens the prod webapp.

PR #597, opened 2018-12-12
Discord bot which posts each time a quest is published

Saumya29:


services/quests/src/bot.tsx
```diff
@@ -0,0 +1,6 @@
+import * as Discord from 'discord.js';
```
toddmedema: This seems like an odd place for this, since `quests` is run on the client. This should be part of the API server new quest publish route, so that we aren't sending a bunch of extra code to the client

toddmedema: (this might also fix the tests failing)

toddmedema:
@Saumya29 ping :)

PR #599, opened 2018-12-14
Fix new quest creation

### LINKED ISSUE #595: Can't create new quests, "cannot access from Google Drive"
Most likely due to the realtime API deprecating to read-only mode

smartin015:
Due to realtime utils implementation, any error with realtime throws an `alert()` and sets `window.location.href = "/"`, which isn't exactly "fail silently".

Tests added as stubs since significant mocking/refactoring would be required to test new quest creation effectively, and prod is broken now.

Fixes #595.

### DIFF
--- a/services/quests/src/actions/Quest.tsx
+++ b/services/quests/src/actions/Quest.tsx
@@ -25,6 +25,19 @@ import {setSnackbar} from './Snackbar';
 const ReactGA = require('react-ga') as any;
 const QueryString = require('query-string');
 
+// Override realtime api error handling, which by default
+// calls alert() and sets window.href to "/" whenever an error occurs.
+realtimeUtils.__proto__.onError = (error: any) => {
+  if (error.type === window.gapi.drive.realtime.ErrorType
+      .TOKEN_REFRESH_REQUIRED) {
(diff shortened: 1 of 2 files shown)


services/quests/src/actions/Quest.tsx
```diff
@@ -25,6 +25,21 @@ import {setSnackbar} from './Snackbar';
...
+// calls alert() and sets window.href to "/" whenever an error occurs.
+if (realtimeUtils) {
+  realtimeUtils.__proto__.onError = (error: any) => {
+    if (error.type === window.gapi.drive.realtime.ErrorType
+        .TOKEN_REFRESH_REQUIRED) {
+      realtimeUtils.authorizer.authorize(() => {
+        console.log('Error, auth refreshed');
+      }, false);
+    } else {
+      console.error(error);
```
toddmedema: I feel like we should also be logging this on our end / to Google Analytics....

smartin015: This is probably going to happen on every new quest until realtime API is perma-deprecated... let's hold off on logging this to GA for now.

Issue #602, opened 2018-12-15
Auto-roll not working, doesn't re-roll each round


toddmedema:
Both in quests and the Custom Combat, the "Auto Roll" function won't reroll between rounds--it spits back the one first set of numbers each and every round (ex. Round 1: 16, 1, 11 ; Round 2+: 16, 1, 11). Good thing to add a unit test for!

PR #610, opened 2018-12-23
create test for new auto-roll each round

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition/issues/602

### DIFF
--- a/services/app/src/components/views/quest/cardtemplates/combat/Actions.test.tsx
+++ b/services/app/src/components/views/quest/cardtemplates/combat/Actions.test.tsx
@@ -196,6 +196,20 @@ describe('Combat actions', () => {
       const {actions} = runTest({});
       expect(actions[2].node.ctx.templates.combat.mostRecentRolls.length).toEqual(3);
     });
+    test('generates new rolls on new rounds', () => {
+      const {actions, store} = runTest({});
+      const firstRolls = actions[2].node.ctx.templates.combat.mostRecentRolls;
+      store.dispatch(handleCombatTimerStop({
+        elapsedMillis: 1000,
+        node: actions[2].node,
+        multiplayer: m.s2p5,
+        seed: '',
+        settings: s.basic
+      }));
+      console.log(store.getActions());
+      const secondRolls = store.getActions()[8].node.ctx.templates.combat.mostRecentRolls;
+      expect(firstRolls).not.toEqual(secondRolls);
+    })
     test('increments the round counter', () => {
       const {actions} = runTest({});
       expect(actions[2].node.ctx.templates.combat.roundCount).toEqual(1);
--- a/services/app/src/components/views/quest/cardtemplates/combat/Actions.test.tsx
+++ b/services/app/src/components/views/quest/cardtemplates/combat/Actions.test.tsx
@@ -194,7 +194,7 @@ describe('Combat actions', () => {
     });
     test('generates rolls according to player count', () => {
       const {actions} = runTest({});
-      expect(actions[2].node.ctx.templates.combat.mostRecentRolls.length).toEqual(3);
+      expect(actions[2].node.ctx.templates.combat.mostRecentRolls.length).toEqual(s.basic.numLocalPlayers);
     });
     test('generates new rolls on new rounds', () => {
       const {actions, store} = runTest({});
@@ -206,7 +206,6 @@ describe('Combat actions', () => {
         seed: '',
         settings: s.basic
       }));
-      console.log(store.getActions());
       const secondRolls = store.getActions()[8].node.ctx.templates.combat.mostRecentRolls;
       expect(firstRolls).not.toEqual(secondRolls);
     })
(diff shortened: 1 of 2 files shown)


services/app/src/components/views/quest/cardtemplates/combat/Actions.test.tsx
```diff
@@ -196,6 +196,20 @@ describe('Combat actions', () => {
...
+      const {actions, store} = runTest({});
+      const firstRolls = actions[2].node.ctx.templates.combat.mostRecentRolls;
+      store.dispatch(handleCombatTimerStop({
+        elapsedMillis: 1000,
+        node: actions[2].node,
+        multiplayer: m.s2p5,
+        seed: '',
+        settings: s.basic
+      }));
+      console.log(store.getActions());
```
smartin015: rm

services/app/src/components/views/quest/cardtemplates/combat/Actions.test.tsx
```diff
@@ -196,6 +196,20 @@ describe('Combat actions', () => {
       const {actions} = runTest({});
       expect(actions[2].node.ctx.templates.combat.mostRecentRolls.length).toEqual(3);
     });
+    test('generates new rolls on new rounds', () => {
```
smartin015: Can you also check that the number of rolls matches the number of adventurers (not players) in a separate test?

services/app/src/components/views/quest/cardtemplates/combat/Actions.tsx
```diff
@@ -357,7 +357,7 @@ export const handleCombatTimerStop = remoteify(function handleCombatTimerStop(a:
   dispatch(audioSet({peakIntensity: 0}));
 
   a.node = a.node.clone();
-  const arng = seedrandom.alea(a.seed);
+  const arng = seedrandom.alea(a.seed + combat.roundCount);
```
smartin015: A regression in this stuff could affect multiplayer syncing - can you add a unit test that the seed is the same for a player that plays through multiple rounds, then syncs multiplayer?

Probably also worth a comment that this is added because the seed is tied to the ParserNode (which remains the same node during combat). 

Alternatively we could make this easier by creating a new node after every action in combat, but that's a much larger change.

services/app/src/components/views/quest/cardtemplates/combat/Actions.test.tsx
toddmedema: That's already captured in the test `generates rolls according to player count`

services/app/src/components/views/quest/cardtemplates/combat/Actions.tsx
toddmedema: Wouldn't creating a new node have the same problem, since it's not a problem with the node but rather that we're passing the same starting seed to a fresh, deterministic random generator?

Am not familiar enough with the multiplayer code to write a test for that in a bounded amount of time :( Are seeds generated new for each node? If so, that's good - but if, say, it's the same seed for the whole session or quest, we're going to have a problem with the rolls being the same for the same round of combat, i.e. the first round of each fight would have the same rolls

smartin015: Every time a new node is constructed, its [context is updated](https://github.com/ExpeditionRPG/expedition/blob/master/shared/parse/Node.tsx#L40) and [a new seed is generated](https://github.com/ExpeditionRPG/expedition/blob/master/shared/parse/Context.tsx#L207) based on the old seed. It can be overridden (e.g. when [cloning a node](https://github.com/ExpeditionRPG/expedition/blob/master/shared/parse/Node.tsx#L58)), but typically every new node has a different seed than the previous node. RNG is done by doing something like `const rng = seedrandom.alea(seed)` at the start of rendering a card, then `rng()` whenever a new value is needed. So by creating a new node when rendering a new combat card, the rng automatically uses a new seed for the new card.

toddmedema: So, I see that shared/Node.tsx generates a new seed on each card (which is tested), and there are tests to make sure that nodes with the same seed are handled deterministically, but I'm not seeing a test that MULTIPLAYER_SYNC handles seeds correctly / seeds persist. Perhaps that would be enough to feel comfortable with this? Since I'm not sure how I'd do that as a whole integration test in a non-fragile manner

smartin015: I think that's what I was asking for in the first comment :) the MULTIPLAYER_SYNC would need to sync through a combat to expose any weirdness in seed caused by this or other combat PRs.

toddmedema:
Ok if I merge this to fix the issue? Unlikely I'll have time to figure out how to test multiplayer any time soon, and this feature is still broken in prod :(

toddmedema:
@smartin015 would you be able to take a crack at writing the MP test for this? We've gotten a few notes from users + from Greg that they miss this feature, would be awesome to be able to include it in the list of app improvements in the "Your copies are shipping!" update

toddmedema:
No longer relevant with https://github.com/ExpeditionRPG/expedition/pull/737

PR #617, opened 2019-01-06
Update style_guide.md

ChrisChinchilla:
Clarified the actual style name and some more details on what that means. Also books follow different style guides :)


toddmedema:
Oh neat, nice find! And thank you for the contribution, if you email me at todd@fabricate.io with the Google email you use to log in to Expedition, I'll get your account credited with your 100 loot points :)

PR #623, opened 2019-01-09
Replace custom quest code with custom quest QDL

### LINKED ISSUE #581: Create "custom combat" quest
The `GM Corner` component is almost exactly the same as the `Tutorials` component, differing only in that it has a special hook to handle custom combats. 

There's also a lot of special-casing that happens based on whether the combat is part of a quest or not: https://github.com/ExpeditionRPG/expedition/search?utf8=%E2%9C%93&q=.custom&type=

Publishing a "custom combat" quest that just loops 'round itself would significantly simplify our code, make it less likely we have regressions in that area, and make it easier to e.g. add multi-round-based XP.

smartin015:
Fixes #581

Adds ~140 lines of config and QDL to remove ~200 lines of custom combat code. This also consolidates the GM corner and Tutorial cards, as they're now both just lists of quests.

image
image
image


toddmedema:
Looking good! Two thoughts:
- should we add the options to start a skill check event in between combats?
- should we label the tiers' buttons with a rough difficulty to help GM's better balance? (i.e. 1 = trivial, 2-3 = easy, 4-5 = medium, 6-7 = hard, 8+ = impossible)

smartin015:
* I'd prefer not to add features in this PR, rather just replace what already exists.
* Seems reasonable, but it wouldn't be easy to do that without repeating it across all buttons. Does that still seem good to you?

toddmedema:
1. Fair
2. Yep, repeating the text on the buttons is fine

PR #626, opened 2019-01-10
Document decisions in the QDL guide

### LINKED ISSUE #531: Add skill checks to QDL docs
https://github.com/ExpeditionRPG/expedition/blob/master/services/quests/docs/qdl_guide.md

Now that The Future is official, we should make skill checks official. Can wait until we've sorted out realtime, just wanted to record it.

smartin015:
Fixes #531

### DIFF
--- a/services/quests/docs/qdl_guide.md
+++ b/services/quests/docs/qdl_guide.md
@@ -336,13 +336,13 @@ To do this, you simply annotate the combat on win outcome. Below, you can see wh
 
 Note: In cases where adventurers won't heal at the end of combat, it's nice to warn them with an instruction node before combat starts so that they can make better choices - for example: `> You're on the run! You won't heal after this encounter.`
 
-### Decision cards
+### Decision cards (The Future expansion)
 
-Decisions allow you to add a more structure element of action to your quest. These can happen both in combat and during roleplay - in fact, they can happen automatically in combat if The Future expansion is enabled!
+Decisions allow you to add a more structure element of action to your quest. These can happen both in combat and during roleplay - in fact, they sometimes happen automatically in combat!
 
 A decision plays out similarly to a round of combat - there's a timed phase where users have a limited amount of time to select the best action to take given some scenario you present.
 
-Decisions can be used without The Future, but they really shine when users have Skill cards that they can use to affect their success or failure when attempting a skill check. Try to use these only in quests where you mark The Future as a requred expansion.
+Decisions can be made when when users have Skill cards that they can use to affect their success or failure when attempting a skill check. If you use decisions, you must mark The Future as a requred expansion.
 
 When out of combat, skill checks look sort of like a special roleplay card:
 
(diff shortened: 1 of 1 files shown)


services/quests/docs/qdl_guide.md
```diff
@@ -310,21 +336,85 @@ To do this, you simply annotate the combat on win outcome. Below, you can see wh
...
 
-### Roleplaying Cards
+Decisions allow you to add a more structure element of action to your quest. These can happen both in combat and during roleplay - in fact, they can happen automatically in combat if The Future expansion is enabled!
 
-Everything except for combat happens in roleplay cards, including showing text and choices to adventurers. The basic syntax for a roleplaying card is:
+A decision plays out similarly to a round of combat - there's a timed phase where users have a limited amount of time to select the best action to take given some scenario you present.
 
-```
-_Card Title_
+Decisions can be used without The Future, but they really shine when users have Skill cards that they can use to affect their success or failure when attempting a skill check. Try to use these only in quests where you mark The Future as a requred expansion.
```
toddmedema: Can they, though? I feel like we would need to make a separate instruction set in the app for if there's a skill check without Skill cards...

services/quests/docs/qdl_guide.md
```diff
@@ -310,21 +336,85 @@ To do this, you simply annotate the combat on win outcome. Below, you can see wh
 
 Note: In cases where adventurers won't heal at the end of combat, it's nice to warn them with an instruction node before combat starts so that they can make better choices - for example: `> You're on the run! You won't heal after this encounter.`
 
+### Decision cards
 
-### Roleplaying Cards
+Decisions allow you to add a more structure element of action to your quest. These can happen both in combat and during roleplay - in fact, they can happen automatically in combat if The Future expansion is enabled!
```
toddmedema: What, not continuing the salsa metaphor? ;)

services/quests/docs/qdl_guide.md
```diff
@@ -310,21 +336,85 @@ To do this, you simply annotate the combat on win outcome. Below, you can see wh
...
+
+* on failure
+
+  Failure!
 ```
 
-Note that you cannot title roleplaying cards `combat` / `Combat`, because that is reserved for combat cards.
+* Notice that there's multiple outcomes - similar to roleplay "choices" - three of these will be shown to users in the timed phase, so there needs to be at least 3 distinct kinds of outcome for users to select from. 
+* Each of these outcomes must specify at least the skill being used - one of knowledge, athletics, or charisma.
+* If you want to get fancy, you can also add a persona alignment ("light" or "dark") that gives players bonus points if they have that particular kind of persona (a type of card from The Horror expansion).
```
toddmedema: (if you reference Persona cards, you must mark your quest as requiring The Horror expansion)

services/quests/docs/qdl_guide.md
smartin015: Fix'd.

services/quests/docs/qdl_guide.md
smartin015: I only included the salsa references to pico their gallos ^^

services/quests/docs/qdl_guide.md
smartin015: Done.

smartin015:
Updated, ptal

PR #629, opened 2019-01-10
add The Future to offline rules

toddmedema:


rules/English.md
```diff
@@ -229,6 +231,9 @@ When you level up, all adventurers may learn a new ability:
 - Add 1 to your ability deck, and place the remaining 2 at the bottom of the deck you drew from.
 - You _may_ choose to discard an ability.
 
+_If playing with The Horror_: Instead of learning an ability, you may also choose to increase your Persona 2 levels.
```
smartin015: I think there needs to be a newline between L234 and L235

PR #630, opened 2019-01-11
Performance

### LINKED ISSUE #252: Find and fix single worst performance bottleneck
_Wednesday May 16, 2018 at 19:08 GMT_
_Originally opened as https://github.com/ExpeditionRPG/expedition-app/issues/687 (2 comment(s))_

----

One specific issue: pretty large lag after clicking buttons (one user specifically reported on iOS)

Ideas:
- try switching away from onTouchTap and to onClick, which modern mobile browsers supposedly support without a delay, but we'll want to test both on mobile web and mobile apps to confirm it still works as expected
  - (if we can't do that) is there a new way to do the touch intercept stuff? We might be subject to the 300ms input delay for some reason...
- streamline components to only need the store data they actually need - https://reactrocket.com/post/react-redux-optimization/
- onClick function takes 300ms on a 4x CPU throttle!
  - looks like render is getting called 5 times, each time takes 30ms. Relevant research: https://gist.github.com/paulirish/5d52fb081b3570c81e3a
- does disabling the CSS transition help? Might be a better way to do that in newer versions of React

Other more general performance optimizations, if it's still slow
- optimize CSS selectors
- disable hot module reloading and look at memory profile
  - disabling sound helps a lot - we could disable sound by default on all apps (not just the web version), but that would be sad :(

smartin015:
Closes #252

- Hint about updates on dialog boxes
- Stop using `<Truncate/>` which causes forced layout and disrupts the render cycle
- Put more stuff in `requestAnimationFrame` so it's not done during render (especially addEventListener which takes 3-5ms per call, and is called ~6x for some UI elements)
- stop re-rendering multitouch canvas 
- Do less event listener registration for UI elements that don't need it
- Stop animating after double-tap


services/app/src/Style.scss
```diff
@@ -1390,6 +1390,24 @@ body.android .dialog .textfield:focus-within {
           display: inline-block;
         }
       }
+      .twoLinesMax {
+         overflow: hidden;
+         text-overflow: ellipsis;
+         display: -webkit-box;
+         -webkit-box-orient: vertical;
+         -webkit-line-clamp: 2; /* number of lines to show */
+         line-height: 1em;        /* fallback */
```
toddmedema: Line height should be 1.2 for optional readability

services/app/src/components/views/QuestPreview.tsx
```diff
@@ -129,7 +129,7 @@ const QuestPreview = (props: Props): JSX.Element => {
         <div>{quest.summary}</div>
         <div className="author">by {quest.author}</div>
         {lastSaved !== null && <div className="summary">Last saved {Moment(lastSaved).fromNow()}</div>}
-        {(quest.ratingcount && quest.ratingcount >= 1) ? <StarRating readOnly={true} value={+ratingAvg} quantity={quest.ratingcount}/> : ''}
+        {(quest.ratingcount && quest.ratingcount >= 1 && quest.ratingcount < 10000) ? <StarRating readOnly={true} value={+ratingAvg} quantity={quest.ratingcount}/> : ''}
```
toddmedema: Err, why is this capped?

smartin015:
Updated, PTAL

PR #643, opened 2019-01-24
Stop timeouts when unmounting splash screen, add tests

### LINKED ISSUE #637: Splash page bug, auto advances on double tap without input
If you double tap the splash page, sometimes it'll skip through the player / mode select screen without user input. Which, besides being annoying, can also break the app, since there's enough time to increase the player count, but it'll still be set to multitouch without you realizing it, which means the first time you hit combat, you'll have to restart the app b/c you won't be able to end combat.

smartin015:
Fixes #637


toddmedema:
(once tests pass, of course)

PR #651, opened 2019-02-10
User badges, scarred lands enemies

### LINKED ISSUE #631: Design returning backer badges & add to account page
1x / 2x / 3x backer (might as well make 4x while we're at it)

Can pull a new field from the DB that Todd will then populate from Kickstarter records.

Probably worth thinking of badges as something a user could have multiple of / designing the DB to support that from the start. At the very least, having 1, 2, and 3x - but also potentially awarding badges for publishing your first quest, writing your first review, playing 10 quests, that sort of thing.

smartin015:
image

Small 3x backer icon (others are similar, just different number of stars):

image

Example account badge view:

image

Closes #631


toddmedema:
Thoughts on implementing the badges as three Boolean columns on the user table, rather than a whole new table + API endpoint? I suspect it would save at least a hundred lines of code, and wouldn't add more "rows" to our Heroku db usage counter

smartin015:
Per #631 if we're expecting users to have many possible badges, it's more future-proofy and results in less schema bloat in our tables. The added code is a consequence of following how things were implemented with `getUserFeedbacks`, which I didn't want to refactor as part of this change.

I also doubt this'll be the feature that puts us over the 10 million row count.

TL;DR: I'd really prefer not to have to implement this twice.

toddmedema:
If you think this implementation is fine, I'm fine with it

toddmedema:
Can't seem to find the approve button on mobile, but lgtm. If you can get it deployed to beta before you leave for work tomorrow and let Gabbie know, she can get it tested tomorrow during the day :)

PR #652, opened 2019-02-18
Disabling combat timer on prepare card switches to non-timer card

smartin015:
Fixes beta test bug: "if you disable timer mid-combat, it becomes a timer of zero and immediately starts counting down into negative numbers with higher damage taken. disabling timer before combat works fine."


smartin015:
Opened #658

PR #665, opened 2019-03-30
Re-fix combat onround error

### LINKED ISSUE #638: Combat on round errors when no next goto
Blocking publishing Mistress Malaise - https://quests.expeditiongame.com/#0BzrQOdaJcH9MWTRBNmNXT1d2UkU

(which is also blocking us publishing a public version of Mistress Malaise that is marked as "Official" - so, at the moment, Mistress Malaise is not available to logged in users)

smartin015:
Fixes #638 (again)

The test passed but prod still had the bug because of a slight inconsistency with establishment of the "root" element. When actually crawling, the first element *within* the `<quest>` tag is actually set as the root element, but the test treated the `<combat>` element as the root. This is important when calling `find()` as it originally searched within the root element to check and see if the IMPLICIT_END event happened inside of combat. It didn't find a node, so failed open.

This change makes the test look more like reality (where `root` is the first node and not the node under test), and the code is now changed to seek up to the `<quest>` tag before calling `find()`.

### DIFF
--- a/services/quests/src/playtest/StatsCrawler.test.tsx
+++ b/services/quests/src/playtest/StatsCrawler.test.tsx
@@ -85,7 +85,7 @@ describe('StatsCrawler', () => {
       expect(Array.from(crawler.getStatsForId('A1').outputs)).toEqual(['IMPLICIT_END']);
     });
 
-    test.only('ignores implicit ends within combat "on round" event triggers', () => {
+    test('ignores implicit ends within combat "on round" event triggers', () => {
       const xml = cheerio.load(`<quest>
         <roleplay data-line="1"><p>prefixing with roleplay sometimes causes errors</p></roleplay>
         <combat title="A1" id="A1" data-line="2">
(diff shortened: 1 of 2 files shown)


services/quests/src/playtest/StatsCrawler.test.tsx
```diff
@@ -85,16 +85,18 @@ describe('StatsCrawler', () => {
       expect(Array.from(crawler.getStatsForId('A1').outputs)).toEqual(['IMPLICIT_END']);
     });
 
-    test('ignores implicit ends within combat "on round" event triggers', () => {
-      const xml = cheerio.load(`<combat title="A1" id="A1" data-line="2">
+    test.only('ignores implicit ends within combat "on round" event triggers', () => {
```
toddmedema: Shouldn't be committed as .only

smartin015: Done.

PR #678, opened 2019-05-15
Scarred lands content set / Enum content sets

### LINKED ISSUE #657: Add "ScarredLands" as content set
Both as a publishing option in QC, and in the app (since it doesn't depend on other expansions, perhaps it should be a checkmark?)

Will require a new DB column, and reviewing the code for all places where other expansion fields are referenced to make sure it gets imported everywhere

smartin015:
Fixes #657

- Content sets are now an `enum` typescript type instead of just a bunch of strings
- Support scarred lands content set
  - new DB column
  - added checkbox to content set selection dialog
  - plumbed scarred lands icon/badge through to quest buttons and quest preview pages
  - updated quest creator to allow publishing with scarred lands set enabled
  - added new search option "compatibility" so that users can check for quests spanning the maximum number of their expansions

Manual tests:

- [x] Icon for displayed quests
- [x] Verify button behavior in quest creator
- [x] Verify content set selection behavior in app
- [x] Verify search behavior in app
- [x] Add column in DB before submission/rollout with default FALSE


toddmedema:
(to both beta and prod)

smartin015:
Already done :) since it's a default field, it's handled fine in the interim.

Issue #422, opened 2018-08-05
Quest crawler bug: not properly detecting undefined variables


toddmedema:
Should error to let writers know that their quest is broken before they publish.... ideally it describes the route in which the undefined was discovered, so that authors know where to look to debug / fix it?

See this PR for when this happens: https://github.com/ExpeditionRPG/expedition/pull/424#pullrequestreview-143948843

PR #707, opened 2019-06-19
Propagate op errors, add default timing metadata

### LINKED ISSUE #698: Bug: Max time validation error
When creating a new quest (and potentially with old quests where min time hasn't been set / the quest hasn't been published yet), if you try to publish, you get a validation error about maxtime... it says that it can't be less than min time, which is weird because the dialog shows min time as 10 minutes. As soon as you select min time (even just picking 10 minutes again), the error no longer occurs.

I suspect what's happening is that min time is being set in the UI, but isn't entered into the validation store until it's picked up by the change handler? (Which does seem a bit weird, since other fields like min/max players don't seem to have the same issue)

smartin015:
Fixes #698
#422

Quest creator previously ignored failures in op node parsing within the body of the quest - this promotes those op errors to real errors when the crawler encounters them, so that quests are more likely to succeed in the wild.

### DIFF
--- a/services/quests/src/playtest/PlaytestCrawler.tsx
+++ b/services/quests/src/playtest/PlaytestCrawler.tsx
@@ -89,9 +89,9 @@ export class PlaytestCrawler extends StatsCrawler {
     }
   }
 
-  protected onWarnings(q: StatsCrawlEntry, warnings: Error[], line: number) {
-    for (const w of warnings) {
-      this.logger.warn(w.toString(), '427', line);
+  protected onErrors(q: StatsCrawlEntry, errors: Error[], line: number) {
+    for (const e of errors) {
+      this.logger.err(e.toString(), '427', line);
     }
   }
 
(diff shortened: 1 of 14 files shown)


services/quests/src/playtest/PlaytestCrawler.tsx
```diff
@@ -89,9 +89,9 @@ export class PlaytestCrawler extends StatsCrawler {
     }
   }
 
-  protected onWarnings(q: StatsCrawlEntry, warnings: Error[], line: number) {
```
toddmedema: This change seems a little risky out of context, since it makes it look like we're turning all warnings into errors. But, I trust you know what you're doing :)

smartin015: It's only used for reporting op errors, and it doesn't actually affect the app's behavior in handling them... should be fine to change it.

I also developed a "playtest runner" that runs the default search, fetches the XML, and runs the playtester on all of them concurrently when you visit `quests.expeditiongame.com/runner.html`... PTAL.

toddmedema: Runner file LGTM, curious how the CI issue gets resolved

PR #713, opened 2019-06-24
Added storage bytes indicator, tests

### LINKED ISSUE #685: Bug: Local storage exceeded errors masked to user
```
QuotaExceededError: Failed to execute 'setItem' on 'Storage': Setting the value of 'SAVED_QUESTS-1PnLDgpmcvCjcdn4fv-qOqXoFKP6BUnfZ-1558901780997' exceeded the quota
```

We should show the users how much local storage space they have, and return a proper error when a new quest can't be saved.

smartin015:
Fixes #685 by displaying available storage space in saved quest section. I don't think we have to worry about this error state often; chrome defaults to 10MB of storage space and most quest are in the 10's of KB range.

image


toddmedema:
For this to be really useful, I think two additional things are needed:
1. Displaying how much space each save takes 
2. Showing an error when you attempt to save and there's not enough space to save... AND / OR since that's a bit late to be useful, a warning when you save that that's the last save you have space for / you'll need to delete other saves to be able to save additional games.

smartin015:
> For this to be really useful, I think two additional things are needed:
> 
> 1. Displaying how much space each save takes

I don't think this is feasible currently... quests take up different amounts of space. We could customize whatever display on space we have for an individual quest, but that becomes much less useful.

> 2. Showing an error when you attempt to save and there's not enough space to save... AND / OR since that's a bit late to be useful, a warning when you save that that's the last save you have space for / you'll need to delete other saves to be able to save additional games.

Verified that storage space errors are shown correctly to be about storage space. Again per #1, we won't know whether or not a user will be out of space until they try to save a quest.

toddmedema:
> > For this to be really useful, I think two additional things are needed:
> > 
> > 1. Displaying how much space each save takes
> 
> I don't think this is feasible currently... quests take up different amounts of space. We could customize whatever display on space we have for an individual quest, but that becomes much less useful.

Ah, so I was thinking that each saved quest listing on this page would show how much space it takes (like the Android app manager lets you see how much storage each app uses)

> > 1. Showing an error when you attempt to save and there's not enough space to save... AND / OR since that's a bit late to be useful, a warning when you save that that's the last save you have space for / you'll need to delete other saves to be able to save additional games.
> 
> Verified that storage space errors are shown correctly to be about storage space. Again per #1, we won't know whether or not a user will be out of space until they try to save a quest.

That's fair. LGTM once #1 is resolved

smartin015:
image

image

Adding file size required changes to the ordering of quest saving and some new state info, PTAL.

Issue #706, opened 2019-06-17
Add Of Wyrms + Giants Booster to App + QC + CC so that we can have a writing contest for it


toddmedema:
Related idea: Have a shared constant with all of the info for each of these card sets, since they're used in several places (Quest Creator publish dialog, App list of content sets & encounters file, Card Creator list of sources)

PR #717, opened 2019-06-30
add wyrms giants

toddmedema:
Fixes https://github.com/ExpeditionRPG/expedition/issues/706

Problem: QC doesn't seem to pick up on it / publish it even if selected...


services/app/src/components/base/QuestButton.tsx
```diff
@@ -46,6 +46,7 @@ export default class QuestButton extends React.Component<Props, {}> {
             {this.props.isOffline && <img className="inline_icon" src="images/offline_small.svg"/>}
             {quest.expansionhorror && <img className="inline_icon" src="images/horror_small.svg"/>}
             {quest.expansionfuture && <img className="inline_icon" src="images/future_small.svg"/>}
+            {quest.expansionwyrmsgiants && <img className="inline_icon" src="images/beast_small.svg"/>}
```
smartin015: We should probably have an actual icon for this...

services/quests/src/components/Dialogs.tsx
```diff
@@ -314,6 +314,13 @@ export class PublishingDialog extends React.Component<PublishingDialogProps, {}>
               onChange={(checked: boolean) => { handleMetadataChange(quest, 'expansionfuture', checked); }}>
             </Checkbox>
           </FormControl>
+          <FormControl className="fullWidth">
+            <Checkbox
+              label="Requires &quot;Of Wyrms &amp; Giants&quot; Booster"
+              value={metadata.get('expansionwyrmsgiants')}
+              onChange={(checked: boolean) => { handleMetadataChange(quest, 'expansionwyrmsgiants', checked); }}>
```
smartin015: I'm a little sad we have a string constant accessor only in this one place. I don't suppose as part of this PR you could refactor `handleMetadataChange` so it takes a Partial<Quest> and then use it to set one or multiple values?

services/app/src/components/base/QuestButton.tsx
toddmedema: Suggestions welcome! Figure this would be ok, though, since the booster itself doesn't have a separate icon anywhere on it, and it's 100% Beast enemies, this seems about as intuitive as using a new / unrelated icon for it

services/quests/src/components/Dialogs.tsx
toddmedema: Yeah :/ I don't like doing big refactors in the same PR as adding functionality, though, since that makes it harder to tell what's a functionality change vs what's a no-op refactor

Issue #718, opened 2019-06-30
Advanced rules reference + glossary - streamline main rules, quick access deeper rules


toddmedema:
Chance to reduce confusion, streamline learning and overall improve play experience by exposing only the general rules / steps at any moment (setup, combat, skill checks, etc), and providing an easy way to look up the details.

- Skill checks: Each roll done by an adventurer above zero health that hasn't yet rolled. This counts as a roleplaying roll.
- Migrating things over from the LTP page, and reference this section of the app on that page - https://expeditiongame.com/begin

Ideally this is both contextually available, and available as a Rules Glossary / Help at any time from the top right menu.

toddmedema:
(Come to think of it, this is basically the opposite of the current ShowHelp setting's implementation - don't show the extended help by default. BUT, also make it really easy to see more help when you need it, i.e. not having to go into settings to toggle something)

PR #719, opened 2019-06-30
streamline rules wordings

toddmedema:
Streamlines combat and skill check rules wordings. Further work could be done with https://github.com/ExpeditionRPG/expedition/issues/718, but that's a larger discussion.

The new combat rules:
Screen Shot 2019-06-30 at 12 09 53
Screen Shot 2019-06-30 at 12 10 04


toddmedema:
The only thing I can think of to make it measurably shorter is #718, since you do still need to have explicit rules somewhere and you can only compress that information so much

PR #725, opened 2019-07-09
Eliminate cycle-prone error reporting

smartin015:
raven-for-redux loops indefinitely when attempting to serialize recursive objects, unlike vanilla JS serialization which detects cycles. This PR filters out ParserNode objects when sending debugging state information to Sentry - these typically contain self-references which cause the app to hang.


services/quests/src/playtest/Runner.tsx
```diff
@@ -83,6 +91,7 @@ function startSearching() {
       console.log('started worker for quest ' + q.id);
       store.dispatch({type: 'PLAYTEST_START', id: q.id, title: q.title, author: q.author, email: q.email});
     }
+    console.log(loadedQuests);
```
toddmedema: Console log

PR #735, opened 2019-07-17
Improve playtest worker, Runner page

### LINKED ISSUE #726: Playtester with incorrect currentCombatTier, :image: support
Errors that aren't actually real errors, but show up via the auto-playtester:

```
{"text":"Error: Failed to evaluate conditional on element: Error: No access to method \"currentCombatTier\" Op: (splitParty == true & _.currentCombatTier() == 3)","type":"error","url":"427","line":226}
```

```
{"text":"[melee] should be on its own line","type":"error","url":"435","line":88}
```

smartin015:
- Playtest worker now has a "visit limit" configurable option that prevents the same node (same data-line) from being seen too many times, which should encourage more broad seeking of different parts of the quest.
- Runner now includes stats, including percentage of "overruns" i.e. where the crawler timed out before crawling the whole quest.
- Default playtest timeout raised to 30 seconds to allow more complicated quests to be crawled more thoroughly.
- Adds currentCombatTier to context.

Fixes #726

### DIFF
--- a/services/app/src/Store.tsx
+++ b/services/app/src/Store.tsx
@@ -18,7 +18,19 @@ export function installStore(createdStore: Redux.Store<AppStateWithHistory>) {
 export function createAppStore(raven: any = null) {
   const middleware = [createMiddleware(getMultiplayerConnection())];
   if (raven) {
-    middleware.push(createRavenMiddleware(raven));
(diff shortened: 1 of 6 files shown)


smartin015:
Yep, I've been reading from the prod DB when testing overruns... Good thought as well on highlighting overrun quests.

PR #749, opened 2019-08-06
1.21.0; fix level-up text

toddmedema:
User-caught bug: "For leveling up after a battle it says "pick one" even if horror or future expansions are not checked"

Fixed that, also took the liberty of removing the "The Horror / The Future" leading text / just kept the icon, makes it look much cleaner and easier to digest

Screen Shot 2019-08-05 at 21 27 53
Screen Shot 2019-08-05 at 21 28 03


smartin015:
I'd suggest removing the colons after the icons, otherwise LGTM when tests are fixed :)

PR #754, opened 2019-08-13
WIP: Allow for serializable quest context; refactor quest saving/loading

smartin015:
Come back later when finished!

This PR makes parser nodes serializable by prohibiting user-defined functions in MathJS. Looking through all published quests over the years, this feature has never been used.

Additional validation will alert users when their quest would be unlikely to load from a save in this fashion.

The old behavior (recreating a node from a path) will still continue to work, and can operate as a fallback if the node or context recreation fails while we're vetting this approach.

This change also opens us up to a lot less buggy networking option - we no longer need to maintain a fragile stack, we can just pass the whole context over the wire and not worry about local state transitions matching remote ones.


smartin015:
This is getting large; will split into multiple PRs so it's easier to understand/review.

smartin015:
Subsumed by other PRs

Issue #742, opened 2019-07-28
Figure out where quest errors are coming from


toddmedema:
Not sure where these are coming from, since we just did a cleanup pass...

- Dragon Wranglers, unsupported node type null around the 2nd combat, tier 4
- Escape from Fallhaven Castle Dungeon, Attempting to 'Next' out of the combat victory screen, on PC, Could not get next node for event "win" - current node tag "null" visible keys []" comparison key {"ctx":"{\"_templateScopeFn\":\"function populateScopeFn(){return(0,o.default)({contentSets:function contentSets(){for(var t=e(),n=t.settings,r=t.multiplayer,i={},o=[].concat((0,a.default)(s.getContentSets(n,r))),u=0;u
- Bounty on Laima Prime, on a combat victory screen (confirmed that this one is not a previous save)
  - crash appears to have moved from the page where it was, to the next page with a new "error!" message at the bottom, but no indication of why it fails or what the error is.
- Quest: On The Pillar of Darkness Part II. Error: Uncaught Error: Could not get next node for event "win" - current node tag "null" visible keys []" comparison key {"ctx":"{\"_templateScopeFn\":\"function populateScopeFn(){return(0,o.default)({contentSets:function contentSets(){for(var t=e(),n=t.settings,r=t.multiplayer,i={},o=[].concat((0,a.default)(s.getContentSets(n,r))),u=0;u
- Quest: Staff of Worldly-Might: Chapter 1. Error: Uncaught Error: Could not find next node. Source: file:///android_asset/www/bundle.js line 12
  - This one was known to be on multiplayer, with two devices / 3 players
-  Iron Fang Raiders. Error: Cannot read property 'wayfinder' of undefined Op: (_.viewcount.wayfinder == 0). Source: shared/parse/context line undefined

smartin015:
Another: 

`Quest: Aramoor Ver. 11. Error: Uncaught Error: Unsupported node type null`

Quest creator link: https://quests.expeditiongame.com/#1MhjHqBwTAEZy-K4rcyW7FRUjG07PPjPc

Quest Line #: 1256

toddmedema:
Backlogging this, as we've deployed numerous fixes in the past few months and the rate of errors seems to have slowed

sciorms:
I encountered this error when updating the tutorial quest but I worked around it. I'll do some testing on this issue to see if I can reliably reproduce it to hopefully squash this once and for all.

PR #756, opened 2019-08-14
Refactor Combat & Decision phases

smartin015:
- Combat & Decision phases are now enums
- Instead of repeated code to generate and set combat template context if not set, rely on Params' `resolveParams` which does the same thing.
- Include combat phase in the combat template context; this allows for a further refactor to centralize the logic that chooses what phase of combat/decision/etc. to render, which is itself a requirement for refactoring the save/sync logic to be able to resume just from a snapshot of node context.


smartin015:
#742

PR #759, opened 2019-08-15
Mandate presence of combat/decision template context, other fixes

smartin015:
- Moved the `Phase` enum definitions to `app/Constants` to prevent circular dependency which wasn't caught by Typescript
- Turn the `combat` and `decision` parts of node context from possibly-undefined to required in the type. Instead of presence of the context values, we can infer what state we're in by the node's element (`decision` vs `combat` vs `roleplay`)
- Remove the `Params` code which really was just a hacky wrapper around the fact that sometimes combat and decision types were missing.
- Move decision phase information from `combat.decisionPhase` to `decision.phase`.


smartin015:
#742

PR #760, opened 2019-08-16
De-phase checkout & multiplayer UI

smartin015:
Completes a long-standing todo to flatten the card state structure for all non-quest cards.

This brings us closer to removing state.card.phase entirely, so we can instead rely on `ParserNode.ctx.templates.(decision|combat).phase`, which will allow us to save/load/sync using just the current node state without having to replay actions.

Confirmed in local testing that the UI still behaves as expected.


smartin015:
#742

PR #761, opened 2019-08-16
(WIP) Refactor: rm card.phase, externalize _history reducer

smartin015:
- Breaks out the actions that operate on `_history` from `CombinedReducers.tsx` to `History.tsx`, for clarity.
- Removes `state.card.phase` attribute, and instead moves all logic handling combat/decision phase to look at `ParserNode.ctx.templates.(combat|decision).phase`. 
- Refactors `toPrevious` action to instead pass a `matchFn` as a way of identifying a matching card in the history.
- Adds an `inCombat` method to `ParserNode` to more easily assess whether the current card should be considered a combat state.

This is the last of a series of refactors in preparation for regenerating the current state of a quest solely from the `ParserNode` and the quest's XML, which is an attempt to fix the major source of errors as in #742.

### DIFF
--- a/services/app/src/actions/ActionTypes.tsx
+++ b/services/app/src/actions/ActionTypes.tsx
@@ -6,8 +6,8 @@ import {ParserNode} from '../components/views/quest/cardtemplates/TemplateTypes'
 import {
   AudioDataState,
   AudioState,
-  CardState,
   CardName,
+  CardState,
   CheckoutState,
   DialogIDType,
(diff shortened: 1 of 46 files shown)


toddmedema:
Lgtm once tests confirmed passing. There is one loose console log in there :) good catch on adding keys to the level up list items

smartin015:
Test passed if you click through to Travis - github must've somehow not gotten the update.

PR #764, opened 2019-08-20
Refactor quest saving/loading

smartin015:
Part of #742. Next step will be to apply this same approach to resolving network diffs in multiplayer.

- Changes quest saving and loading from "re-trace a path from the start" to "re-load a snapshot of the context at the node where the save was created". This is much more performant and much more resistant to bugs that may arise due to changes in how nodes are evaluated and rendered.
- This new quest saving/loading strategy means we must no longer support user-defined functions within op nodes. However, this is completely unused in all quests in the prod DB, so I don't think that's any great loss.
- Part of this change affects how our provided functions are bound; instead of being bound at creation time (and every time the node is copied), they are instead bound at op evaluation time. This may actually improve performance, but it also simplifies sites where contexts are constructed and results in a much simpler `ctx.scope` overall (no need to pass around a template scope generator any more!)
- The "legacy" load code remains to allow for users to still load their old quests; however, we should aim to stop supporting this path after a few months.

### DIFF
--- a/services/app/src/actions/SavedQuests.test.tsx
+++ b/services/app/src/actions/SavedQuests.test.tsx
@@ -11,7 +11,17 @@ describe('SavedQuest actions', () => {
   const STORED_QUEST_TS = 67890;
 
   beforeEach((done) => {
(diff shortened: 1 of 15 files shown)


toddmedema:
Oh no, tests failed :(

smartin015:
Was actually a flaky test - ran `yarn test` and `build-all` locally, both pass.

Issue #770, opened 2019-08-26
Bug: "Expedition" is listed as an expansion in search


toddmedema:
Also, the "Expansion" header should be pluralized

toddmedema:
Looks to be fixed

PR #771, opened 2019-08-27
checkbox to see official quests; closes #605

### LINKED ISSUE #605: Filter checkbox: "Only show official quests"
Enabling it should only show quests with official = true (defaults to off)

Can have it on the bottom, just above the "Also show my private quests"

toddmedema:
Screen Shot 2019-08-26 at 21 39 52
Screen Shot 2019-08-26 at 21 40 16


toddmedema:
@smartin015 applied diff model, PTAL as it involved some casting stuff that I verified works, but want to make sure the code make sense to other folks / that's a good way of doing it

Also added one line of related code to not show base game as expansion; closes #770

PR #822, opened 2020-01-06
Handle api downtime

### LINKED ISSUE #815: App handle API downtime better
It should retry twice, and then indicate on the splash screen and search that the server is down, and not show the login button, instead, please try again in a few minutes - if the issue persists, you can contact support at contact@fabricate.io

Lordtien:
- Adding the 'fetch-retry' module.
- Retrying twice at an interval of a second in case of Server Error, followed by an announcement with a retry message.
- Updating 'isLatestVersion' as an optional in 'ServerStatusState' interface.
- Adding a test for added changes.

Fixes #815

### DIFF
--- a/services/app/src/actions/ServerStatus.tsx
+++ b/services/app/src/actions/ServerStatus.tsx
@@ -20,7 +20,6 @@ export function fetchServerStatus(log: any = logEvent) {
     .then((data: FetchServerStatusResponse) => {
       dispatch(handleServerStatus(data));
     }).catch((error: Error) => {
-      // Don't alert user - it's not important to them if this fails
       dispatch(setServerStatus({
         announcement: {
           open: true,
--- a/services/app/src/actions/ServerStatus.tsx
+++ b/services/app/src/actions/ServerStatus.tsx
@@ -21,12 +21,14 @@ export function fetchServerStatus(log: any = logEvent) {
       dispatch(handleServerStatus(data));
     }).catch((error: Error) => {
       const isLatestAppVersion = false;
+      const serverOffline = true;
       dispatch(setServerStatus({
         announcement: {
           open: true,
           message: 'Please try again in a few minutes. If the issue persists, you can contact support at contact@fabricate.io',
         },
         isLatestAppVersion,
+        serverOffline,
       }));
       log('error', 'status_fetch_err', {label: error});
     });
--- a/services/app/src/actions/ServerStatus.tsx
+++ b/services/app/src/actions/ServerStatus.tsx
@@ -20,15 +20,13 @@ export function fetchServerStatus(log: any = logEvent) {
     .then((data: FetchServerStatusResponse) => {
       dispatch(handleServerStatus(data));
     }).catch((error: Error) => {
-      const isLatestAppVersion = false;
-      const serverOffline = true;
       dispatch(setServerStatus({
         announcement: {
           open: true,
           message: 'Please try again in a few minutes. If the issue persists, you can contact support at contact@fabricate.io',
         },
-        isLatestAppVersion,
-        serverOffline,
+        isLatestAppVersion: false,
+        serverOffline: true,
       }));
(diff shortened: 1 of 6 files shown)


toddmedema:
@Lordtien looks like there's a typescript error with your change (you can see it by clicking on the "details" next to the red X in the "All checks have failed" section:

```
ERROR in [at-loader] ../app/src/components/views/ModeSelectContainer.tsx:11:3 
    TS2322: Type '{ isLatestAppVersion: boolean | undefined; settings: SettingsType; user: UserState; multiplayer: ...' is not assignable to type 'StateProps'.
  Types of property 'isLatestAppVersion' are incompatible.
    Type 'boolean | undefined' is not assignable to type 'boolean'.
      Type 'undefined' is not assignable to type 'boolean'.
```

Lordtien:
> @Lordtien looks like there's a typescript error with your change (you can see it by clicking on the "details" next to the red X in the "All checks have failed" section:
> 
> ```
> ERROR in [at-loader] ../app/src/components/views/ModeSelectContainer.tsx:11:3 
>     TS2322: Type '{ isLatestAppVersion: boolean | undefined; settings: SettingsType; user: UserState; multiplayer: ...' is not assignable to type 'StateProps'.
>   Types of property 'isLatestAppVersion' are incompatible.
>     Type 'boolean | undefined' is not assignable to type 'boolean'.
>       Type 'undefined' is not assignable to type 'boolean'.
> ```

@toddmedema While dispatching, we need to pass ```isLatestAppVersion```. Is it a good idea to pass that as ```false```, since if we have a server error, we cannot be sure?

toddmedema:
So, thinking about it more, I acfhaklg5 like your original approach of leaving it as undefined - because we really don't know, since there was a server error.

Having it be undefined just means that you need to check to see if it's defined before it's referenced, i.e. `if (typeof yourvar !== 'undefined')`

Lordtien:
So, to me, my initial method seemed like a hack, since...
1) Taking more than expected out of a boolean, i.e. isLatestAppVersion
2) Might break code at other places.

Suggestions:
1) Keep it false, but adds another thing to remember.
2) Add a new property, if we expect to use this in the future, like, isConnected or isServerOnline.

For both, I'll be exploring the places we're using this as of now.

Lordtien:
In `initialServerStatusState`, we do put in `isLatestAppVersion` as `false`, since it's unknown at that point in time. I think we can still do the same so as to keep the logic consistent.
On the other hand, if we go with keeping it `undefined`, then we need to make changes for the new possible value as well as comment it at the definition so that it can be read.
IMO, the second idea still sounds like a hack as we're pushing the application of a boolean and it adds to the "Things to remember".
@toddmedema Need your comments before proceeding further.

toddmedema:
Good brainstorming! Let's do `serverOffline` as a new field, so that we can succinctly check if the server is definitely not available via `if (serverOffline)`

Lordtien:
Alright. Adding it as an optional field with `false` as a default value. Once the check fails, or maybe if the server doesn't respond mid-way, we can set it to `true`.

Keeping `isLatestAppVersion` as `false` for this case.

services/app/src/actions/ServerStatus.tsx
```diff
@@ -10,15 +11,27 @@ import {FetchServerStatusResponse, ServerStatusSetAction} from './ActionTypes';
...
+    return fetch(AUTH_SETTINGS.URL_BASE + '/announcements', {
+      retries: 2,
+      retryDelay: 1000,
+    })
+    .then(handleFetchErrors)
+    .then((response: Response) => response.json())
+    .then((data: FetchServerStatusResponse) => {
+      dispatch(handleServerStatus(data));
+    }).catch((error: Error) => {
+      const isLatestAppVersion = false;
```
toddmedema: Style note: Since these values are simple and never change, we should just define them as part of the object, rather than separate variables

PR #823, opened 2020-01-29
Bug 7 players

### LINKED ISSUE #811: Bug: Can select 7 ppl with multitouch
And then no quests show up in search

Obv don't want to cap it behind the scenes without telling them, since that'd be confusing ... perhaps we show a modal on the next screen saying "you tried 7+ people, but we capped it to 6 since Expedition only supports 1-6 players"

Lordtien:
In the case of more than 6 players, a dialog saying "Too many players" open up.
Peek 2020-01-29 17-05

Issue: Once the dialog is canceled, the splash screen sticks to the last state and requires an additional touch to reset. I need help there.

Fixes #811

### DIFF
--- a/services/app/src/Init.tsx
+++ b/services/app/src/Init.tsx
@@ -95,7 +95,7 @@ function setupDevice() {
     // https://stackoverflow.com/a/43502958/1332186
     if (/Android/.test(navigator.appVersion)) {
       window.addEventListener('resize', () => {
-        if (document.activeElement !== null) {
+        if (document.activeElement) {
           if (document.activeElement.tagName === 'INPUT' || document.activeElement.tagName === 'TEXTAREA') {
             document.activeElement.scrollIntoView();
           }
--- a/services/app/src/Init.tsx
+++ b/services/app/src/Init.tsx
@@ -95,8 +95,10 @@ function setupDevice() {
     // https://stackoverflow.com/a/43502958/1332186
     if (/Android/.test(navigator.appVersion)) {
       window.addEventListener('resize', () => {
-        if (document.activeElement.tagName === 'INPUT' || document.activeElement.tagName === 'TEXTAREA') {
-          document.activeElement.scrollIntoView();
+        if (document.activeElement !== null) {
+          if (document.activeElement.tagName === 'INPUT' || document.activeElement.tagName === 'TEXTAREA') {
+            document.activeElement.scrollIntoView();
+          }
         }
       });
     }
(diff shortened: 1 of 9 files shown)


Lordtien:
@toddmedema As the PR description mentions, I need to reset the splash screen post dialog cancellation. I cannot find a way to do that. Can you provide some info?

toddmedema:
You'd probably want to reset state to its initial value: https://github.com/ExpeditionRPG/expedition/blob/master/services/app/src/components/views/SplashScreen.tsx#L27

Lordtien:
Update: Made the change. Here's a GIF (Player limit set to 0 for testing):
Peek 2020-01-29 17-05

services/app/src/Init.tsx
```diff
@@ -95,8 +95,10 @@ function setupDevice() {
     // https://stackoverflow.com/a/43502958/1332186
     if (/Android/.test(navigator.appVersion)) {
       window.addEventListener('resize', () => {
-        if (document.activeElement.tagName === 'INPUT' || document.activeElement.tagName === 'TEXTAREA') {
-          document.activeElement.scrollIntoView();
+        if (document.activeElement !== null) {
```
toddmedema: In JS, you can also abbreviate this by just checking if (document.activeElement), since if it's null, it'll be falsey :)

services/app/src/components/base/Dialogs.test.tsx
```diff
@@ -245,4 +245,27 @@ describe('Dialogs', () => {
       expect(e.find('#play').length).toEqual(0);
     });
   });
+
+  describe('TooManyPlayersDialog', () => {
```
toddmedema: Thank you for adding a test!

services/app/src/components/base/Dialogs.tsx
```diff
@@ -208,6 +208,35 @@ export class ExpansionSelectDialog extends React.Component<ExpansionSelectDialog
...
 }
 
+export class TooManyPlayersDialog<T extends BaseDialogProps> extends React.Component<T, {}> {
+  protected title: string;
+  protected content: JSX.Element;
+
+  constructor(props: T) {
+    super(props);
+    this.title = 'Too many players!';
+    this.content = <p>Expedition can only be played by a maximum of 6 players!</p>;
```
toddmedema: I'd replace the exclamation mark in contents with a period, since there's already one in the title and too many exclamation marks comes off a bit strong

services/app/src/Init.tsx
Lordtien: Got it.

services/app/src/components/base/Dialogs.tsx
Lordtien: Alrighty. Thanks.

Issue #750, opened 2019-08-06
LTP quest: Make the market choice meaningful


toddmedema:
Creates a bad first impression of the Expedition system if the first choice you make doesn't matter

sciorms:
@toddmedema  I have some thoughts regarding this issue that could be useful to talk about before I submit a PR against this issue. After reviewing the intro quest's xml, I see two things that can be easily change to add more impact to this encounter:

1) There are three branches during the market scene of the quest but two of the branches are basically identical with slightly different flavor text. I think an easy way to handle this would be to only show two options on "The Market" scene which are either attempt to intimidate the faeries verbally or use force to subdue them and then transition from "Intimidation Tactics"  or "Force" to "Attempted Arrest". The "Attempted Arrest" scene may need to be modified as well based on the success of either branch. 

2) There seems to be no way to actually intimidate the faeries in the market. I think you could add a roll prompt here where a 15+ successfully intimidates them and they surrender peacefully. I believe a 15+ would be a good roll because new guards would not have tons of influence within the town but they may have some minimal influence because they are guards nonetheless. If this roll fails then the fairies attack the group and start combat.

If you think this is a good potential solution I could write this up and submit a PR for it. Let me know what your thoughts are.

toddmedema:
That sounds reasonable! I will say, the reason we did that is that the LTP needs combat to teach you about combat - so maybe the intimidation still has combat, but makes it easier? (and lets the user know that it's easier, i.e. you deal +1 damage this combat because you intimidated them)

sciorms:
@toddmedema this can probably be marked as fixed per PR #839

PR #839, opened 2020-05-19
 changing flow of intro quest

sciorms:
@toddmedema Here is my first pass at issue #750. 

Let me know what feedback you have. I went through the readme to install the local testing suite using yarn but ran into a few errors while running yarn test so I was not able to test this locally.

### DIFF
--- a/services/app/src/quests/learning_to_adventure.xml
+++ b/services/app/src/quests/learning_to_adventure.xml
@@ -1,215 +1,216 @@
 <quest title="Learning to Adventure" data-line="0">
-  <roleplay icon="adventurer" title="Your adventure begins" data-line="2">
-    <p>This quest teaches you how to play the base Expedition game. To learn expansion mechanics (such as The Horror's <i>Personas</i> and The Future's <i>Skills</i>), you'll want to play the other Learning to Adventure quests after this one.</p>
-    <instruction if="_.numAdventurers() &gt; 1">
-      <p>One player should read this page out loud, then pass the device to the next player.</p>
-    </instruction>
-    <p>"Alright, recruits, everyone line up!" The grizzled town guardsman marches up and down the line, inspecting the uniforms and proper posture of the new guards.</p>
-    <p>You and your friends steal glances between one another as the old guardsman approaches you, standing as tall as you can to mask the first-day jitters.</p>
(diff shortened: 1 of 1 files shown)


services/app/src/quests/learning_to_adventure.xml
```diff
@@ -47,15 +47,27 @@
...
           <roleplay icon="fae" title="Intimidation Tactics" data-line="60">
-            <p>{{intimidated = 1}}</p>
             <p>You march up to them, a hand on the hilt of your newly acquired sword.</p>
             <p>"As guardsmen of the town watch, we demand you stop disturbing the peace!"</p>
-            <p>...</p>
-            <p>They ignore you. You try again. "STOP!" you yell as loud as you can.</p>
-            <p>Nothing. You'll have to try a different strategy.</p>
+            <choice text="Rolled above 10">
+                <p>{{intimidated = 1}}</p>
+                <p>You successfully intimidate the fairies</p>
```
toddmedema, line 54: Perhaps this should go to the combat directly? Seems weird to have it say that intimidation succeeded, and then go back to the "try another option" screen

services/app/src/quests/learning_to_adventure.xml
```diff
@@ -127,6 +133,7 @@
   <combat data-line="153">
     <e>Will o' the Wisp</e>
     <e>Magic Mushroom</e>
+    <e if="intimidated == 1">+2 to all attacks made during this combat.</e>
```
toddmedema, line 136: I would explain to the user that this is because their intimidation succeeded

sciorms:
@toddmedema pushed a new version per your feedback. I changed where the intimidation bonus shows up in the combat text and modified the flow between the market scene and the magic fight scene to be a bit more coherent. Let me know your thoughts.

services/app/src/quests/learning_to_adventure.xml
sciorms: I relocated this text to the combat tutorial screen and modified the text to mention that the bonus is because of the intimidation.

services/app/src/quests/learning_to_adventure.xml
sciorms: I modified this outcome to flow directly into the subue/force scene as a transition into the arrest scene which then flows into combat.

sciorms:
fixes #750

PR #840, opened 2020-05-22
resolving spelling errors in app quests

sciorms:
Quick PR to resolve a few spelling errors in existing quests in app. @toddmedema

### DIFF
--- a/services/app/src/quests/a_shiny_score.xml
+++ b/services/app/src/quests/a_shiny_score.xml
@@ -7,7 +7,7 @@
     <p>Put those answers aside for the moment. They will be selected and chosen by the players as the quest continues.</p>
   </roleplay>
   <roleplay title="Introduction" data-line="14">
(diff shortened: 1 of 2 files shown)


sciorms:
I updated the base LTP quest and published/unpublished it per your instructions however there's a weird inconsistency with "A Shiny Score". 

From the files changed in this commit, you can see that the original had "accomodations" in the text which I updated to "accommodations", however when I open up the shared quest in QC "accommodations" is spelled right in the text. I'm not sure where the discrepancy stems from but should I proceed to publish/unpublish the quest in QC even though it is technically already in parity with its associated xml file?

toddmedema:
Oh interesting - sounds like Greg might have fixed that after the last time we synced it into the app.

I'd go ahead and sync it just in case there've been any other changes since the last time we synced...

Perhaps we should open an issue for an easier way to sync quests into the app :D

sciorms:
Good thought! I published and unpublished both quests so they should be in sync now. All good to merge?

toddmedema:
Just to clarify, did you also do the XML copy-paste part? Don't see any new commits to this PR since I gave you access :)

sciorms:
I added the updated xml file from Shiny Score and verified that both appear as expected. Feel free to double check :)

PR #872, opened 2021-11-09
Migrate to google sheets CSV since sheets API changed

### LINKED ISSUE #871: Update card creator to use sheets API v4
https://cloud.google.com/blog/products/g-suite/migrate-your-apps-use-latest-sheets-api

smartin015:
With the advent of Sheets API V4, the easiest way to fetch the card sheets is to fetch the sheet as a (public) CSV link. This PR fixes #871 by replacing Tabletop.JS (now deprecated) with PapaParse for CSV parsing. 

Additionally:

* Moved sheet keys/uids into `Constants.tsx` (previously in `services/cards/src/reducers/Filters.tsx`). Declared a new interface type so the structure is regular.
* Custom flow now shows two prompts: one for selecting the type of card and the second for linking to the specific sheet the cards are in. This is a worse user experience, but a necessary one as there is no easy way to fetch all tabs of the same google sheet at once any more :(

### DIFF
--- a/services/cards/src/Constants.tsx
+++ b/services/cards/src/Constants.tsx
@@ -16,7 +16,7 @@ export const SHEETS: CardSource[] = [
       Translations: '2111992856',
       Helper: '987926921',
       Adventurer: '558719643',
-      Ability: '0',
+      Ability: '0', // actually the sheet ID, not a placeholder
       Encounter: '1555320979',
       Loot: '1510752952',
     },
@@ -37,7 +37,7 @@ export const SHEETS: CardSource[] = [
     key: '2PACX-1vQcPWt9MO7ZDtLaR1teWcuNxoRHH4Q1Yf4xLpyv3xSEqSOLbpPVj_aB5Emh7sOnZ3hMIwvA9LJn76pZ',
     sheets: {
       Adventurer: '558719643',
-      Ability: '0',
+      Ability: '0', // actually the sheet ID, not a placeholder
       Encounter: '1555320979',
       Loot: '1510752952',
       Skill: '1013479922',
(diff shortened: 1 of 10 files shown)


services/cards/src/Constants.tsx
```diff
@@ -1,3 +1,63 @@
...
+
+export const SHEETS: CardSource[] = [
+  {
+    name: 'Expedition',
+    key: '2PACX-1vQ0Hx1RcDqkc3pn7ZyKEiqdDXm4miEgDeozJ9amJrxz2QzAH8AOaf6rqaLJL4G3fcqCDIYDKdDqsuax',
+    sheets: {
+      Translations: '2111992856',
+      Helper: '987926921',
+      Adventurer: '558719643',
+      Ability: '0',
```
toddmedema, line 19: Should this be 0? We should have an abilities sheet for the base game (and for the future below)

services/cards/src/actions/Filters.tsx
```diff
@@ -12,13 +12,11 @@ const qs = require('qs');
...
-    if (name === 'source' && value === 'Custom') {
+    let cardstype = null;
+    if (name === 'source' && value === 'custom') {
       // TODO validate URL or ID, otherwise notify user + abort
-      value = window.prompt('Please enter your card sheet publish URL (see "?" in the top right for help)', '') as string;
-      if (value.indexOf('/e/') !== -1) {
-        return alert('Please use the URL of the Google Doc, not the publish link');
-      }
-      value = 'Custom:' + value.replace('https://docs.google.com/spreadsheets/d/', '').split('/')[0];
+      cardstype = window.prompt('Please enter a sheet type (Ability, Adventurer, Encounter, Helper, Loot, Persona, Skill)');
```
toddmedema, line 18: 1. Is it possible to get the tab name from the API so that we don't have to ask the user?
2. Whatever the final flow is (incl how users can generate the new URLs, if that's changed), can you update the docs too? https://github.com/ExpeditionRPG/expedition/blob/master/services/cards/CARD-CREATION.md

smartin015: 1. The API doesn't include the sheet type anywhere - so we can't infer it without some fuzzy matching or modifying the structure of the sheets - both of which I'd prefer not to do in an "unbreak" PR
2. Done

services/cards/src/Constants.tsx
smartin015: 0 is the ID of the Ability sheet. Added a clarifying comment

